Issue #17, opened 2025-05-20
Declarative Arrays


Vezio:
### Component

Setup/Standup

### Describe the bug

When running the setup script, namely, `./setup/standup.sh -s 0-5`, a user that is on an older version of `bash` will not have support for declarative arrays.

### Steps to reproduce

`./setup/standup.sh -s 0-5`

```
==> Tue May 20 16:37:29 EDT 2025 - ./setup/standup.sh - 🔍 Checking if "llmd-vezio" is prepared.
/REDACTED/llm-d-benchmark/setup/env.sh: line 278: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/REDACTED/llm-d-benchmark/setup/env.sh: line 280: llama: unbound variable
/REDACTED/llm-d-benchmark/setup/env.sh: line 278: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/REDACTED/llm-d-benchmark/setup/env.sh: line 280: llama: unbound variable
/REDACTED/llm-d-benchmark/setup/env.sh: line 278: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
/REDACTED/llm-d-benchmark/setup/env.sh: line 280: llama: unbound variable
```

### Additional context or screenshots

N/A

maugustosilva:
Fixed on PR #22

PR #22, opened 2025-05-20
[Setup/standup] fix: remove dependency on associative arrays #17

maugustosilva:


diegocastanibm (approved):
This modification is necessary to be able to run the automation scripts with old versions of bash (like the ones in mac).

PR #42, opened 2025-06-11
Add inference-perf and fixes for using non-openshift environment

achandrasekar:
This change adds the following changes:
- Add inference-perf as another benchmark harness that can be run
- Add a profile that can be run with inference-perf
- Add a scenario that can be run with inference-perf on GKE
- Fix minor issues that were encountered


maugustosilva:
@achandrasekar Thank you for this, tested locally, worked (for `fmperf`).

There are some additional cleanups, but we can do this on a subsequent PR.

PR #93, opened 2025-06-30
Add probes to standalone deployment, add envar for storage request and limit

namasl:
- Add a startup probe to the standalone deployment, which has a very tolerant `failureThreshold` to allow for both quick- and slow-to-load models. Setting a long `initialDelaySeconds` results in unnecessary waiting time. By probing every 30 seconds and allowing for many failures allows for quick detection whenever the model is ready, over a wide timeframe (up to `periodSeconds` * `failureThreshold`). After the startup probe passes, the liveness and readiness take over.
- Add environment variables for ephemeral storage request and limit. The hard-coded values are too small for large models.


setup/steps/06_deploy_vllm_standalone_models.sh
```diff
@@ -53,25 +53,36 @@ spec:
...
+            port: ${LLMDBENCH_VLLM_COMMON_INFERENCE_PORT}
+          failureThreshold: 3
           periodSeconds: 5
         resources:
           limits:
             cpu: "${LLMDBENCH_VLLM_COMMON_CPU_NR}"
             memory: ${LLMDBENCH_VLLM_COMMON_CPU_MEM}
             $(echo "$LLMDBENCH_VLLM_COMMON_ACCELERATOR_RESOURCE: \"${LLMDBENCH_VLLM_COMMON_ACCELERATOR_NR}\"")
-            ephemeral-storage: "20Gi"
+            ephemeral-storage: ${LLMDBENCH_VLLM_STANDALONE_EPHEMERAL_STORAGE_LIMIT}
```
kalantar, line 80: I notice that for other resources, the limit and request are set to the same value. Why is this one treated differently?

namasl: I could set these to the same variable, I don't see any reason for it to matter.

maugustosilva (approved):

PR #94, opened 2025-07-01
handle existing conda install on home or other dir

deanlorenz:
- If conda already exists then should not modify `.bashrc` or assume it is on `/opt`. use `conda info` instead.
  **note:** was checked on Linux, should verify mac path returned (line 39).
- typo in line 31 -- announce instead of echo was used to update .bashrc


maugustosilva:
Tested on a Mac, works as advertised

maugustosilva (approved):

PR #123, opened 2025-07-10
[Run] feat: add support for `vllm` benchmarks as a harness

maugustosilva:
A new harness type, `vllm-benchmark` can be specified. The rendered yaml file is, internally, converted into a list CLI options, with the parameter `executable` specifying the load generator (e.g., `benchmark_serving.py`)

During this process, a small refactor on `llm-d-benchmark.sh` made it simpler to call with different harnesses (and workloads) when the parameter `-d` is used on `run.sh`

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update; \
     rsync \
     patch \
     curl \
+    yq \
     && apt-get clean && rm -rf /var/cache/apt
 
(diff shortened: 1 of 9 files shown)


build/Dockerfile
```diff
@@ -45,22 +46,68 @@ list = yes\n\
...
+    python3 setup.py install
+
+ARG INFERENCE_PERF_REPO=https://github.com/kubernetes-sigs/inference-perf.git
+ARG INFERENCE_PERF_BRANCH=main
+RUN git clone --branch ${INFERENCE_PERF_BRANCH} ${INFERENCE_PERF_REPO}
+RUN cd inference-perf; pip install .
+
+ARG VLLM_BENCHMARK_REPO=https://github.com/vllm-project/vllm.git
+ARG VLLM_BENCHMARK_BRANCH=main
+RUN git clone --branch ${VLLM_BENCHMARK_BRANCH} ${VLLM_BENCHMARK_REPO}
```
namasl, line 65: We may want to tie these repositories to a particular commit

PR #133, opened 2025-07-12
Additional fixes for "parameterless" CLI options on vllm-benchmark

maugustosilva:
Additionally added an example workload (vllm-benchmark) which allows "parameter sweep"


maugustosilva:
A quick example where "parameter sweep" is used.

```
for i in 1 2; do; export LLMDBENCH_RUN_EXPERIMENT_PARAMETER_MAX_CONCURRENCY=$i; export LLMDBENCH_RUN_EXPERIMENT_ID=$i; ./setup/run.sh -m llama-3b -w random_1k_concurrent_10-1_ISL-OSL; done
```

PR #141, opened 2025-07-17
Add preprocess logic to run before standalone vllm server starts

manoelmarques:
Add preprocess logic before the standalone vllm server starts.

The preprocess logic is divided in two steps:
1. bash commands to install necessary libraries for some of the load formats
2. python script. The python script is contained in a configmap created automatically and loaded into the pod.

More than one preprocess script can be used. It can be configured as follows:
```
export LLMDBENCH_VLLM_STANDALONE_PREPROCESS="source /workload/preprocesses/standalone-preprocess.sh;/workload/preprocesses/standalone-preprocess.py"
```


setup/steps/06_deploy_vllm_standalone_models.sh
```diff
@@ -38,11 +61,42 @@ spec:
       - name: vllm-standalone-$(model_attribute $model label)
         image: $(get_image ${LLMDBENCH_VLLM_STANDALONE_IMAGE_REGISTRY} ${LLMDBENCH_VLLM_STANDALONE_IMAGE_REPO} ${LLMDBENCH_VLLM_STANDALONE_IMAGE_TAG})
         imagePullPolicy: Always
-        command: ["/bin/sh", "-c"]
+        command:
+        - /bin/bash
         args:
-        - >
-          $(render_string $LLMDBENCH_VLLM_STANDALONE_ARGS $model)
+        - "-c"
```
maugustosilva, line 67: Question: wouldn't you prefer to have this script baked directly on the image?

workload/preprocesses/standalone-preprocess.py
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python3
```
maugustosilva, line 1: Similar situation... we can build it on the image.

setup/steps/06_deploy_vllm_standalone_models.sh
maugustosilva: In my view, once all the "pre-processing" binaries/executables are inside the image, one should be able to do something like

```
LLMDBENCH_VLLM_STANDALONE_ARGS="standalone-preprocess.py;___vllm____serve____REPLACE_MODEL____--enable-sleep-mode____--load-format____REPLACE_ENV_LLMDBENCH_VLLM_STANDALONE_VLLM_LOAD_FORMAT____--port____REPLACE_ENV_LLMDBENCH_VLLM_COMMON_INFERENCE_PORT____--max-model-len____REPLACE_ENV_LLMDBENCH_VLLM_COMMON_MAX_MODEL_LEN____--disable-log-requests____--gpu-memory-utilization____REPLACE_ENV_LLMDBENCH_VLLM_COMMON_ACCELERATOR_MEM_UTIL____--tensor-parallel-size____REPLACE_ENV_LLMDBENCH_VLLM_COMMON_ACCELERATOR_NR"
```

maugustosilva:
PLEASE DISREGARD MY PREVIOUS COMMENTS :-) Any attempt to bake these binaries inside a custom `vllm` image is a terrible terrible idea. I failed to realize we were talking about the "vanilla" `vllm` image, and **not** the `llm-d-benchmark` image.

maugustosilva (approved):

PR #151, opened 2025-07-20
End to end feat

ryan-mangeno:
high level script to go through setup, run, teardown, in tandem with https://github.com/llm-d/llm-d-benchmark/pull/135, it utilizes the sweeping script, very much a rough draft as of now...

### DIFF
--- a/setup/e2e.py
+++ b/setup/e2e.py
@@ -69,7 +69,6 @@ def handle_sweep(args) -> None:
 
 
 def handle_batch(args) -> None:
-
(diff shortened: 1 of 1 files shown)


maugustosilva:
While I do agree that converting our coding from pure `bash` to `python` is a net positive, this should be done in a top-to-bottom approach. This means converting individual steps from (e.g., 04) from `bash` to `python`, using `pykube` in place of `oc/kubectl` and only then grouping all of these under a "master" `python` script.

PR #158, opened 2025-07-21
[Setup] fix: Don't try to deploy gateway provider in case deployment method is standalone

maugustosilva:


Vezio:
> Where does `LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_MODELSERVICE_ACTIVE` get set?

Looks like it's in `setup/env.sh`

```
for mt in standalone deployer modelservice; do
  is_env=$(echo $LLMDBENCH_DEPLOY_METHODS | grep $mt || true)
  if [[ -z $is_env ]]; then
    export LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_$(echo $mt | tr '[:lower:]' '[:upper:]')_ACTIVE=0
  else
    export LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_$(echo $mt | tr '[:lower:]' '[:upper:]')_ACTIVE=1
  fi
done
```

namasl (approved):

PR #160, opened 2025-07-22
Install python again after changing repo/branch

deanlorenz:
While `llm-d-benchmark.sh` checks out the right branch if different from the default, it should also reinstall the python packages.

*note:* verify that the installs of the default repos works as intended. Currently, some do not install for the checked-out branch.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -85,11 +85,11 @@ COPY analysis/vllm-benchmark-analyze_results.sh /usr/local/bin/vllm-benchmark-an
 COPY analysis/guidellm-analyze_results.sh /usr/local/bin/guidellm-analyze_results.sh
 
 RUN echo "#!/usr/bin/env bash" > /usr/local/bin/llm-d-benchmark.sh; echo -e "\
-
+\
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -120,6 +120,23 @@ if [[ -d \$LLMDBENCH_RUN_EXPERIMENT_HARNESS_DIR ]]; then \n\
     pushd /workspace/\$LLMDBENCH_RUN_EXPERIMENT_HARNESS_DIR\n\
   fi\n\
   git checkout \$LLMDBENCH_HARNESS_GIT_BRANCH\n\
+  case \${1} in\n\
```
maugustosilva, line 123: Better use `${LLMDBENCH_RUN_EXPERIMENT_HARNESS_DIR}` instead of `${1}`

deanlorenz: DONE.

*Question:* `${LLMDBENCH_RUN_EXPERIMENT_HARNESS_DIR}` may match `${1}*`, but the git directory names are hardcoded at the top with the initial git clone. In which scenario could it be different than `${1}`?

PR #165, opened 2025-07-22
converted step 04 to py

ryan-mangeno:
made conversion for step04.sh in setup to python, havent fully tested, waiting for cluster to come back up. In anticipation of functions from functions.sh in pr 159 to be used, this file with need to have the updated function calls instead of using subprocess.run.

### DIFF
--- a/setup/steps/04_ensure_model_namespace_prepared.sh
+++ b/setup/steps/04_ensure_model_namespace_prepared.sh
@@ -102,3 +102,5 @@ EOF
 
   return 0
 }
+
+main


setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+        exec_command(f'announce \"Failed to check storage class and affinity\"')
+        sys.exit(1)
+
+    #  loop through models and call the installer script
+    models = [model.strip() for model in model_list_str.split(',') if model.strip()]
+
+    for model in models:
+        
+        delete_job_cmd = f"\"{kcmd} --namespace \"{vllm_namespace}\" delete job download-model --ignore-not-found\" {dry_run} {verbose}"
+        if exec_command(command=f"llmdbench_execute_cmd \"{delete_job_cmd}")[0]:
```
maugustosilva, line 68: This should be implemented with `pykube`

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+
+        model_artifact_uri = f"pvc://{vllm_namespace}/models/{download_model}"
+        
+        # in prev bash script ... PROTOCOL=${MODEL_ARTIFACT_URI%%://*} is used, and splits by :// 
+        # then pvc_and_model_path is the path after protocol, so they can be asigned with the same split
+        protocol, pvc_and_model_path = model_artifact_uri.split('://') # protocol var unused but exists in prev script
+        pvc_name, model_path = pvc_and_model_path.split('/')
+
+
+        exec_command(f'create_namespace \"{kcmd}\" \"{vllm_namespace}\"')
```
maugustosilva, line 82: ditto

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+
+        exec_command(f'create_or_update_hf_secret \
+                    \"{kcmd}\" \"{vllm_namespace}\" \
+                    \"{vllm_hf_token_name}\" \
+                    \"{vllm_hf_token_key}\" \
+                    \"{hf_token}')
+
+
+        exec_command(f'validate_and_create_pvc \
```
maugustosilva, line 91: ditto

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+                    \"{kcmd}\" \
+                    \"{vllm_namespace}\" \
+                    \"{download_model}\" \
+                    \"{pvc_name}\" \
+                    \"{vllm_storage_size}\" \
+                    \"{vllm_storage_class}\"')
+        
+        exec_command(f'launch_download_job \
```
maugustosilva, line 99: ditto

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+                    \"{kcmd}\" \
+                    \"{vllm_namespace}\" \
+                    \"{vllm_hf_token_name}\" \
+                    \"{download_model}\" \
+                    \"{model_path}\" \
+                    \"{pvc_name}\"')
+        
+        exec_command(f'wait_for_download_job \
```
maugustosilva, line 107: ditto

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,186 @@
...
+        if is_openshift:
+            exec_command(f'llmdbench_execute_cmd \
+                    \"{kcmd}\" \
+                    adm policy add-scc-to-user anyuid \
+                    -z \"{vllm_service_account}\" \
+                    -n \"{vllm_namespace}\" \
+                    \"{dry_run}\" \
+                    \"{verbose}\" 1 1 1')
+            
+            exec_command(f'llmdbench_execute_cmd \
```
maugustosilva, line 124: ditto

ryan-mangeno:
do you want me to abstract the stuff that is the same across multiple exec_command calls?

setup/steps/04_ensure_model_namespace_prepared.py
ryan-mangeno: when I debug what kcmd is set to it is oc, will pykube call kubectl or will it handle this?

maugustosilva: Yes, `pykube` is `python` native `Kubernetes` client. It can be used to managed `Openshift` clusters in the very same way.

ryan-mangeno: ok will do

ryan-mangeno:
@maugustosilva would you rather I convert some of those functions in functions.sh to python and use those?

maugustosilva:
@ryan-mangeno absolutely. Any conversion from `bash` to `python` should start by adding a `functions.py` side-by-side with `functions.sh`

ryan-mangeno:
Hello, made some large additions, converted the commands that were exported from env.sh that were used in the step04 sh into python and just called those in my step04.py script. also allowed ability to pick up python scripts from the steps dir from standup.sh

ryan-mangeno:
also, should I maybe replace announce with ```logging.basicConfig(
    level=logging.INFO,
    format='%(asctime)s - %(levelname)s - %(message)s'
)
logger = logging.getLogger(__name__)```

maugustosilva:
This looks significantly better now. Still going through the PR, but it looks very good so far.

setup/steps/py-steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,160 @@
...
+def main():
+
+    os.environ['CURRENT_STEP_NAME'] =  os.path.splitext(os.path.basename(__file__))[0]
+
+    ev = {}
+    for key in dict(os.environ).keys():
+        if 'LLMDBENCH_' in key:
+            ev.update({key.split('LLMDBENCH_')[1].lower():os.environ.get(key)})
+
+    vllm_namespace =        ev['vllm_common_namespace']
```
Vezio, line 42: A strong suggestion would be to just use the `env` object as it is, rather than manually unpacking each entity into each of their local respective variables. 

For example:

instead of `kcmd = ev['control_kcmd']` and then using `kcmd` throughout, you can merely use `ev['control_kcmd']`. That should clean up all of these declarations here.

ryan-mangeno: just added :)

maugustosilva (approved):

kalantar:
I fetched the PR and modfied `setup/standup.sh` to just print the step that will be executed: 
```
% git diff
diff --git a/setup/standup.sh b/setup/standup.sh
index c83a999..2885268 100755
--- a/setup/standup.sh
+++ b/setup/standup.sh
@@ -180,7 +180,8 @@ for step in ${LLMDBENCH_STEP_LIST//,/ }; do
   then
     step=$(printf %02d $step)
   fi
-  run_step "$step"
+  announce "🚀 Running step ${step}"
+  # run_step "$step"
 done
```

This is the list of steps executed. We should be replacing the .sh step w/ the .py step. Not just executing at the end.
```
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 00_ensure_llm-d-infra.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 01_ensure_local_conda.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 02_ensure_gateway_provider.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 03_ensure_user_workload_monitoring_configuration.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 04_ensure_model_namespace_prepared.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 05_ensure_harness_namespace_prepared.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 06_deploy_vllm_standalone_models.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 07_deploy_gaie.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 08_deploy_via_modelservice.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 09_smoketest.sh
==> Fri Aug  1 18:03:46 EDT 2025 - setup/standup.sh - 🚀 Running step 04_ensure_model_namespace_prepared.py
```

Vezio:
@ryan-mangeno based on the output - there's a brief bug in the `standalone.sh` - you're appropriately checking the execution of `.sh` or `.py` scripts - but you're actually executing `04_ensure_model_namespace_prepared.sh` first (as illustrated in @kalantar output) and then you're _re-executing_ the python converted version at the end. 

You should remove - or archive (realistically Git tracks this so if we needed to go back to it we can always go to a previous commit...so just remove the script entirely) - the initial `bash` implementation for this sake so that the `standalone.sh` doesn't execute it twice, especially out of order.

Be sure to rerun a few times - you may have some hidden bugs in the python implementation since the shell script implementation was accordingly being run first so it seems based on @kalantar output. Good catch.

setup/standup.sh
```diff
@@ -134,6 +134,37 @@ export LLMDBENCH_CONTROL_CLI_OPTS_PROCESSED=1
...
+    local step_nr=$(echo $step_id | cut -d '_' -f 1)
+    export LLMDBENCH_CURRENT_STEP=${step_nr}
+    announce "=== Running step: $step_id ==="
+    if [[ $LLMDBENCH_CONTROL_DRY_RUN -eq 1 ]]; then
+      echo -e "[DRY RUN] $script_path\n"
+    fi
+
+    if [[ $script_path == *.sh ]]; then
+      source $script_path
+    elif [[ $script_path == *.py ]]; then
```
Vezio, line 156: Ensure that the order is preserved in running step `04` - and remove the `shell` implementation of `04`.

Vezio (changes requested):

ryan-mangeno:
is there a reason why there is a run_step function in both functions.sh and standalone? I removed it and added the new version to functions.sh

setup/standup.sh
ryan-mangeno: 👍

kalantar:
Required packages not installed (`setup\install_deps.sh`?):
```
==> Mon Aug  4 12:12:07 EDT 2025 - setup/standup.sh - === Running step: 04_ensure_model_namespace_prepared.py ===
Traceback (most recent call last):
  File "/Users/kalantar/projects/go.workspace/src/github.com/llm-d/llm-d-benchmark/setup/steps/04_ensure_model_namespace_prepared.py", line 8, in <module>
    import pykube
ModuleNotFoundError: No module named 'pykube'
```

ryan-mangeno:
adding installation for deps in install_deps.sh

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -0,0 +1,139 @@
...
+        asyncio.run(wait_for_job(
+            job_name="download-model",
+            namespace=ev["vllm_common_namespace"],
+            timeout=ev["vllm_common_pvc_download_timeout"],
+        ))
+
+    # possibly needs to be impliment, is not very simple with pykube
+    #if is_openshift:
+    #    add_scc_to_service_account(api, "anyuid", vllm_service_account, vllm_namespace, dry_run)
+    #    add_scc_to_service_account(api, "privileged", vllm_service_account, vllm_namespace, dry_run)
```
kalantar, line 108: In the past I have found this is necessary for vllm to function correctly. It seems to have worked without this. However, do we know for certain that it will work for all newer vllm images? I tried it with `ghcr.io/llm-d/llm-d-routing-sidecar:v0.2.0`.

ryan-mangeno:
> I was able to successfully test this PR. It seems to have successfully create the PVC and downloaded the model. I am concerned about the lack of the `adm policy add-scc-to-user` steps. However, in my test, vllm started successfully.

im going to try and impliment this  ill see what i can find

Vezio (approved):

PR #168, opened 2025-07-23
[Setup] feat: add support for (vllm) command-line arguments and pod environment variables on files

maugustosilva:
Additionally, add a new scenario ("precise-prefix-cache-aware") with corresponding new presets


setup/functions.sh
```diff
@@ -233,43 +233,57 @@ function add_annotations {
...
+    for envvar in ${LLMDBENCH_VLLM_COMMON_ENVVARS_TO_YAML//,/ }; do
+      output=$output"REPLACE_NEWLINEREPLACE_SPACESN- name: $(echo ${envvar} | $LLMDBENCH_CONTROL_SCMD -e 's^LLMDBENCH_VLLM_STANDALONE_^^g')REPLACE_NEWLINEREPLACE_SPACESVvalue: \"${!envvar}\""
+    done
 
-  if [[ $LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_MODELSERVICE_ACTIVE -eq 1 ]]; then
-    local spacen="      "
-    local spacev="        "
-  fi
+    if [[ $LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_STANDALONE_ACTIVE -eq 1 ]]; then
+      local spacen="        "
```
Vezio, line 252: Is there a better way to do this instead of with spaces - I was going to suggest this before hand as well. I'm ok with it for now - but 🤷

namasl: What about creating a JSON artifact as a single line? It may be aesthetically ugly, but still valid.

PR #178, opened 2025-07-24
kubeasync addition to remove time.wait(30)

ryan-mangeno:
using kube async python library to remove a time.wait(30) hardcoded in fmperf benchmark script in function that waits for eval job to be done, also removing hardcoded timeout checking. It uses a stream that can be continuously checked asynchronously provided by the kube aync api with ```w = k8s_async_watch.Watch()
        async with w.stream( ...
```

### DIFF
--- a/workload/harnesses/fmperf-llm-d-benchmark.py
+++ b/workload/harnesses/fmperf-llm-d-benchmark.py
@@ -76,17 +76,18 @@ def update_workload_config(workload_spec, env_vars):
 
     return workload_spec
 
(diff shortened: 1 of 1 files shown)


maugustosilva (approved):

ryan-mangeno:
I reworked function sig and removed logging directly in the wait for job function since the same wait for job function is implimented in this pr https://github.com/llm-d/llm-d-benchmark/pull/165, and this should end up importing that function from function.py idealy to avoid repeat code.

Issue #162, opened 2025-07-22
Standardized benchmarking report


namasl:
### Component

Harnesses

### Desired use case or feature

Due to the desire for `llm-d-benchmark` to support a wide array of harnesses, each with their own unique output format, adding a post-processing step to each harness in order to convert the results into a uniform format will greatly simplify downstream operations. Analysis, plotting, and other tooling that utilizes benchmarking results will only have a single data format that needs to be supported.

### Proposed solution

- [x] Create a Python class defining the data schema ([Pydantic](https://docs.pydantic.dev/latest/))
- [x] Implement a post-processing step for each supported harness to convert result data into the standard output format
  - [x] vLLM benchmark
  - [x] Inference Perf
  - [x] fmperf
  - [x] GuideLLM
- [x] Update analysis to use standardized format
- [x] Add documentation on standardized format

We can utilize the proposed format at https://github.com/kubernetes-sigs/inference-perf/issues/124

namasl:
All items completed

PR #185, opened 2025-07-29
Add report definition

namasl:
First part of #162, creates a definition for report format, including validation (importing data not adhering to schema or having invalid entries will be rejected with corresponding errors) and JSON schema generation.

This follows the proposal at https://github.com/kubernetes-sigs/inference-perf/issues/124

### DIFF
--- a/workload/report/schema.py
+++ b/workload/report/schema.py
@@ -1,0 +1,460 @@
+from enum import StrEnum, auto
+import json
+from typing import Optional, Any
+
+from pydantic import BaseModel, model_validator
+import yaml
+
+
+class Parallelism(BaseModel):
+    """Accelerator parallelism details."""
+
+    dp: int = 1
+    """Data parallelism level."""
+    tp: int = 1
+    """Tensor parallelism level."""
+    pp: int = 1
+    """Pipeline parallelism level."""
+    ep: int = 1
+    """Expert parallelism level."""
+
+
+class HostAccelerator(BaseModel):
+    """Host accelerator details."""
+
+    model: str
+    """Accelerator model."""
+    memory: float | int
+    """Amount of memory in one accelerator, in GB."""
+    count: int
+    """Number of accelerators."""
+    parallelism: Optional[Parallelism] = None
+    """Parallelism configuration used."""
+    metadata: Optional[Any] = None
+
+
+class Host(BaseModel):
+    """Host hardware details."""
+
+    accelerator: list[HostAccelerator]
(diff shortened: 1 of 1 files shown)


workload/report/schema.py
```diff
@@ -0,0 +1,460 @@
...
+    type: Optional[str] = None
+    args: Optional[dict[str, Any]] = None
+    metadata: Optional[Any] = None
+
+
+class Scenario(BaseModel):
+    """System configuration and workload details for benchmark run."""
+
+    description: Optional[str] = None
+    host: Optional[Host] = None
```
achandrasekar, line 103: Do we want to provide a way to specify multiple hosts as well when multiple replicas are run?

workload/report/schema.py
```diff
@@ -0,0 +1,460 @@
...
+    """Number of accelerators."""
+    parallelism: Optional[Parallelism] = None
+    """Parallelism configuration used."""
+    metadata: Optional[Any] = None
+
+
+class Host(BaseModel):
+    """Host hardware details."""
+
+    accelerator: list[HostAccelerator]
```
achandrasekar, line 39: How do we specify prefill vs decode servers? Would it be to use metadata to add that information?

achandrasekar (commented):
Thanks for sending this out! This will be super useful. Looks good to me overall - have some minor comments / clarifications.

workload/report/schema.py
maugustosilva: Yes, we need multiple hosts.

workload/report/schema.py
```diff
@@ -0,0 +1,460 @@
...
+    args: dict[str, Any]
+    metadata: Optional[Any] = None
+
+
+class Platform(BaseModel):
+    """Software platform details endompassing all inference engines."""
+
+    engine: list[EngineDetails]
+    """Details on inference engines, list corresponds 1:1 with scenario.host.accelerator."""
+    metadata: Optional[Any] = None
```
maugustosilva, line 57: Would the "presets" used by `GAIE` be stored (in native `yaml` format directly under metadata?

workload/report/schema.py
namasl: I added the field `scenario.host.type`, which is a list of types that can be one of `prefill`, `decode`, or `replica`. `prefill`/`decode` are for dis-aggregated setups, otherwise `replica` is used.

See the commented out demo code at the very bottom for an example YAML.

PR #196, opened 2025-07-30
[Run] feat: ensure all harnesses work properly and output the same data

maugustosilva:
With this PR all harnesses, whether python or bash, will add the used workload profile (an `yaml` file) and a `stdout.log` to the results directory.

Also, fixed a longstanding bug with the moving of the `LLMDBENCH_CONTROL_WORK_DIR` when `./setup/standup.sh` was called with all the steps

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -32,7 +32,7 @@ WORKDIR /workspace
 
 # Install harnesses
 
-# Required by fmperf
+# Required by our fmperf benchmark harness
 RUN pip install kubernetes_asyncio
 
 ARG FM_PERF_REPO=https://github.com/fmperf-project/fmperf.git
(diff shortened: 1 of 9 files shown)


build/Dockerfile
```diff
@@ -31,6 +31,10 @@ sed -i 's^\-e^^' /etc/rsyncd.conf
 WORKDIR /workspace
 
 # Install harnesses
+
+# Required by fmperf
+RUN pip install kubernetes_asyncio
```
namasl, line 36: I would remove this after fmperf install, this way if/when fmperf updates their `requirements.txt` to properly reflect what is actually required, in case they specify particular versions, those versions will get installed and this `pip install` will be passive.

maugustosilva: This is not from `fmperf`, but rather from our own "fmperf client"

maugustosilva: Pushed a fix clarifying the need for it

PR #205, opened 2025-07-31
Add universal report conversion step to vLLM benchmark harness

namasl:
This is the next step in implementing #162, which incorporates a post-processing step to the vLLM benchmark harness creating a report in the standardized benchmarking format.


maugustosilva:
This is very very very welcome and needed. Looks ok to me, LGTM and improve later.

PR #207, opened 2025-07-31
Initial support for multiple models

kalantar:
This provides support to deploy multiple models (even the same model) to the same namespace all sharing the same gateway.  Note that models deployed in different namespaces still require their own gateway. This change should be addressed separately.

Environment variable `LLMDBENCH_DEPLOY_MODEL_LIST` can be defined to be a comma separated list of models of the form `model:model_id`. For example,
```shell
export LLMDBENCH_DEPLOY_MODEL_LIST="facebook/opt-125m:fb,facebook/opt-125m:facebook"
```
describes two copies of the model `facebook/opt-125m` should be deployed with different identifiers.
If the `model_id` is not present, it defaults to `model`.

`model_id` should be unique (per gateway). If different models are deployed and no `model_id` is provided, the defaults will be unique.

One helmfile per model is created. Separate values files for gaie and modelservice are created for each.

Inference requests should include the (lower case of) model_id in the URL path. That is, of the form: `http://gateway_route/model_id/v1/models`

If just one model is deployed, it is not necessary to include the model_id in the URL. This should prevent current use cases from breaking.

This PR also updates `get_model_name_from_pod` to hardcode LLMDBENCH_CONTROL_VERBOSE to 0 in the computation of `response`. Otherwise, the response is not json. Perhaps `llmdbench_execute_cmd` is not appropriate?


setup/functions.sh
```diff
@@ -28,14 +28,18 @@ function model_attribute {
   # Do not use associative arrays. Not supported by MacOS with older bash versions
 
   case "$model" in
-    "llama-1b") local model=meta-llama/Llama-3.2-1B-Instruct ;;
-    "llama-3b") local model=meta-llama/Llama-3.2-3B-Instruct ;;
-    "llama-8b") local model=meta-llama/Llama-3.1-8B-Instruct ;;
-    "llama-70b") local model=meta-llama/Llama-3.1-70B-Instruct ;;
-    "llama-17b") local model=meta-llama/Llama-4-Scout-17B-16E-Instruct ;;
+    "llama-1b") local model=meta-llama/Llama-3.2-1B-Instruct:llama-1b ;;
```
Vezio, line 31: Would it be easier and a bit more future proof to take these in as one single argument instead? 

Let the user specify a huggingface model string such as `meta-llama/Llama-3.2-1B-Instruct:llama-1b` and then let `huggingface` perform the handling to check if it's a valid model, and redirect the stdout/err back to the user of llm-d-benchmark?

Unless the motivation here is to hard limit the user to only a subset of the total models available on huggingface, then in that case, I'm ok with this implementation.

setup/steps/02_ensure_gateway_provider.sh
```diff
@@ -3,17 +3,6 @@ source ${LLMDBENCH_CONTROL_DIR}/env.sh
 
 if [[ $LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_MODELSERVICE_ACTIVE -eq 1 ]]; then
 
-  # make sure llm-d-modelservice helm repo is available
```
jgchn, line 6: If the `helm repo add llm-d-modelservice` is now deleted, how is benchmark supposed to add or upgrade the repo?

setup/functions.sh
Vezio: Pretty good

maugustosilva:
We have a few conflicts to solve, but other than that, LGTM.

kalantar:
Fixed `setup/teardown.sh`  as well. Tested `./run.sh` using `llama-1b` model.

maugustosilva (approved):

Issue #146, opened 2025-07-17
Spyre Integration for Standalone Model Deployment


Vezio:
### Component

Setup/Standup

### Desired use case or feature

Allow the user to select an accelerator they wish to deploy a specific model. For the purposes of this issue, we will enable IBM Spyre model deployment for our _standalone_ specification as a first pass.

### Proposed solution

- Enhance Step `06_deploy_vllm_standalone_models.sh` such that it will deploy a selected accelerator profile

PR #209, opened 2025-08-01
standalone support for spyre

ryan-mangeno:
added standalone support for spyre and utilized scenario from @Vezio


setup/functions.sh
```diff
@@ -397,6 +406,11 @@ function add_command_line_options {
   else
     if [[ $LLMDBENCH_CURRENT_STEP == "06" ]]; then
       echo "  - |"
+      if [[ "${LLMDBENCH_VLLM_COMMON_AFFINITY}" == *"aiu"* ]]; then
+        local indent=$(printf '%*s' 12 '')
```
Vezio, line 410: Thank you for the `12` character indent via printf ;)

setup/functions.sh
```diff
@@ -397,6 +406,11 @@ function add_command_line_options {
...
+        echo "${indent}source /etc/profile.d/ibm-aiu-setup.sh"
```
Vezio, line 411: Good catch here - this is critical for spyre - otherwise everything will be completely and utterly broken.

setup/env.sh
```diff
@@ -155,6 +155,13 @@ export LLMDBENCH_CONTROL_WAIT_TIMEOUT=${LLMDBENCH_CONTROL_WAIT_TIMEOUT:-900}
 export LLMDBENCH_CONTROL_CHECK_CLUSTER_AUTHORIZATIONS=${LLMDBENCH_CONTROL_CHECK_CLUSTER_AUTHORIZATIONS:-0}
 export LLMDBENCH_CONTROL_RESOURCE_LIST=deployment,httproute,route,service,gateway,gatewayparameters,inferencepool,inferencemodel,cm,ing,pod,job
 
+
+if [[ "${LLMDBENCH_VLLM_COMMON_AFFINITY}" == *"aiu"* ]]; then
```
Vezio, line 159: Pretty good - may want to be case insensitive here - but good for now.

Vezio (changes requested):
@ryan-mangeno seems to have tested this configured a bunch of times locally - but I believe there are a few hazards we'd want to be aware of related to the mobile service steps after.

Let's hold off on merging until we address those next week with @maugustosilva and team.

Requesting Changing as a form of blocking PR...

ryan-mangeno:
> @ryan-mangeno seems to have tested this configured a bunch of times locally - but I believe there are a few hazards we'd want to be aware of related to the mobile service steps after.
> 
> Let's hold off on merging until we address those next week with @maugustosilva and team.
> 
> Requesting Changing as a form of blocking PR...

sounds good! thanks for checking it out!

setup/env.sh
ryan-mangeno: ill add that thanks 👍

setup/functions.sh
ryan-mangeno: :)))

ryan-mangeno:
- regarding https://github.com/llm-d/llm-d-benchmark/issues/146

ryan-mangeno:
it now works with most stuff being in scenario file now, also the image on the pr doesnt seem to work, but there is an vllm spyre image I could use that allowed this to work, but it was hardcoded and was formatted like a hashed image.

Vezio (approved):

PR #232, opened 2025-08-07
update fqdn

kalantar:
Fixes a defect in how fqdn is computed in the loop. Subsequent executions were appending the namespace again. This is needed when run loops over multiple models.


maugustosilva (approved):

maugustosilva:
Confirmed that it works, LGTM

PR #240, opened 2025-08-11
Fix for (full) CI/CD

maugustosilva:


.github/workflows/benchmark1.yaml
```diff
@@ -124,7 +124,7 @@ jobs:
       - name: E2E target cloud (modelservice, fmperf)
         env:
           LLMDBENCH_HF_TOKEN: ${{ secrets.LLMDBENCH_HF_TOKEN }}
-        run: ./setup/e2e.sh -c cicd -t modelservice --deep -l fmperf -w sanity_short-input
+        run: ./setup/e2e.sh -c cicd -t modelservice --deep -l fmperf -w sanity_short-input.yaml
```
jgchn, line 127: Why is `.yaml` required here but not in the two edits before this?

maugustosilva: Not "required" but actually "allowed". The idea is to test both cases, with and without `.yaml`

jgchn: Got it makes sense!

jgchn (approved):

PR #243, opened 2025-08-12
Remove model aliases

namasl:
The feature of model aliases obfuscates the exact model used and generally adds other complications. This PR removes the unnecessary feature.


maugustosilva (approved):

maugustosilva:
@jgchn but we can revisit it on a different PR.

PR #247, opened 2025-08-12
[Setup] chore: switch step 04 to python implementation

maugustosilva:
Some additional fixes were required, in particular, auto-detection of "default" storage class.

Step 01 was also switched to the python implementation

Also cleaned up further the model attribute detection

Added `GitPython` as a dependency in `install_deps.sh`

Finally, added the "wait for creation" step on `standalone`


kalantar:
When I tried to run this it failed. `model_attribute` (shell version) has changed. The python version was not changed.  Something like this is needed: 
```
% git diff
diff --git a/setup/functions.py b/setup/functions.py
index 87a4780..840aeee 100644
--- a/setup/functions.py
+++ b/setup/functions.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import subprocess
 import inspect
 import pykube
+import hashlib
 from pykube.exceptions import PyKubeError

 import yaml
@@ -408,9 +409,16 @@ async def wait_for_job(job_name, namespace, timeout=7200):

 def model_attribute(model: str, attribute: str) -> str:

+    model, modelid = model.split(':', 1) if ':' in model else (model, model)
+
     #  split the model name into provider and rest
     provider, model_part = model.split('/', 1) if '/' in model else ("", model)

+    hash_object = hashlib.sha256()
+    hash_object.update(modelid.encode('utf-8'))
+    digest = hash_object.hexdigest()
+    modelid_label = f"{provider[:8]}-{digest[:8]}-{model_part[-8:]}"
+
     # create a list of components from the model part
     # equiv  to: tr '[:upper:]' '[:lower:]' | sed -e 's^qwen^qwen-^g' -e 's^-^\n^g'
     model_components_str = model_part.lower().replace("qwen", "qwen-")
@@ -453,6 +461,8 @@ def model_attribute(model: str, attribute: str) -> str:
     # storing all attributes in a dictionary
     attributes = {
         "model": model,
+        "modelid": modelid,
+        "modelid_label": modelid_label,
         "provider": provider,
         "type": type_str,
         "parameters": parameters,
```

maugustosilva:
That is a good question. In my view the plan is to retire the `bash` implementations ASAP.

maugustosilva:
> Not sure what's happening, if I do dry run: `LLMDBENCH_CONTROL_STEP_04_IMPLEMENTATION=py ./setup/standup.sh -s 4 -v -n`, it takes a long time and I get:
> 
> ```
> 2025-08-12 17:27:15,942 - INFO - ---> would have executed the command "oc apply -n llmd-kalantar-1 -f /Users/kalantar/tmp/benchmark-pokprod-llmd-kalantar-1/setup/yamls/04_download_pod_job.yaml"
> 2025-08-12 17:27:15,943 - INFO - Waiting for job download-model to complete...
> 2025-08-12 17:28:16,252 - INFO - Error occured while waiting for job download-model : Response payload is not completed: <TransferEncodingError: 400, message='Not enough data to satisfy transfer length header.'>
> ```
> 
> This does not actually create anything in the cluster so approving anyway.
> 
> It seems to work fine when not a dry run.

Thanks for testing this... I didn't try dry-run with step 04. Let me investigate a little bit.

kalantar:
I have no trouble with dry-run with the latest update. lgtm.

PR #251, opened 2025-08-13
Convert step 03 to python

yossiovadia:
sample output: 

`(.venv) ➜  llm-d-benchmark git:(convert-step-03-to-python) ✗ LLMDBENCH_CONTROL_DEPLOY_IS_OPENSHIFT=1 LLMDBENCH_CONTROL_WORK_DIR=/tmp/test_step03 LLMDBENCH_CONTROL_DRY_RUN=1 LLMDBENCH_CONTROL_VERBOSE=1 python3 setup/steps/03_ensure_user_workload_monitoring_configuration.py
2025-08-12 18:18:23,472 - INFO - DRY RUN enabled. No actual changes will be made.
2025-08-12 18:18:23,472 - INFO - 🔍 Checking for OpenShift user workload monitoring enablement...
2025-08-12 18:18:23,472 - INFO - ---> would write ConfigMap YAML to /tmp/test_step03/setup/yamls/03_ensure_user_workload_monitoring_configuration_cluster-monitoring-config_configmap.yaml
2025-08-12 18:18:23,473 - INFO - YAML content would be:
apiVersion: v1
data:
  config.yaml: 'enableUserWorkload: true'
kind: ConfigMap
metadata:
  name: cluster-monitoring-config
  namespace: openshift-monitoring

2025-08-12 18:18:23,473 - INFO - ---> would have executed the command "kubectl apply -f /tmp/test_step03/setup/yamls/03_ensure_user_workload_monitoring_configuration_cluster-monitoring-config_configmap.yaml"
2025-08-12 18:18:23,473 - INFO - ✅ OpenShift user workload monitoring enabled`

### DIFF
--- a/setup/steps/03_ensure_user_workload_monitoring_configuration.py
+++ b/setup/steps/03_ensure_user_workload_monitoring_configuration.py
@@ -1,0 +1,198 @@
+import os
+import sys
+import yaml
+from pathlib import Path
+
+# Add project root to path for imports
+current_file = Path(__file__).resolve()
+project_root = current_file.parents[1]
+sys.path.insert(0, str(project_root))
+
+try:
+    from functions import announce, llmdbench_execute_cmd
+except ImportError as e:
+    # Fallback for when dependencies are not available
+    print(f"Warning: Could not import required modules: {e}")
+    print("This script requires the llm-d environment to be properly set up.")
+    print("Please run: ./setup/install_deps.sh")
+    sys.exit(1)
+
+
+def create_monitoring_configmap() -> dict:
+    """
+    Create OpenShift monitoring ConfigMap using native Python dict structure.
+    
+    Returns:
+        dict: ConfigMap structure for enabling user workload monitoring
+    """
+    return {
+        'apiVersion': 'v1',
+        'kind': 'ConfigMap',
+        'metadata': {
+            'name': 'cluster-monitoring-config',
+            'namespace': 'openshift-monitoring'
+        },
+        'data': {
+            'config.yaml': 'enableUserWorkload: true\n'
+        }
+    }
+
+
+def write_configmap_yaml(configmap: dict, output_path: Path, dry_run: bool, verbose: bool) -> bool:
+    """
+    Write ConfigMap to YAML file using Python yaml library.
+    
+    Args:
+        configmap: ConfigMap dictionary structure
+        output_path: Path where to write the YAML file
+        dry_run: If True, only print what would be written
+        verbose: If True, print detailed output
+    
+    Returns:
+        bool: True if successful, False otherwise
+    """
+    if dry_run:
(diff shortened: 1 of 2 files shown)


setup/steps/03_ensure_user_workload_monitoring_configuration.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def main():
+    """Main function following the pattern from other Python steps"""
+    
+    # Set current step name for logging/tracking
+    os.environ["CURRENT_STEP_NAME"] = os.path.splitext(os.path.basename(__file__))[0]
+    
+    # Parse environment variables into ev dictionary (following established pattern)
+    ev = {}
```
maugustosilva, line 170: Given this code gets repeated for all the steps, wouldn't it be better to add it to a function?

```
    ev = {}
    for key, value in os.environ.items():
        if "LLMDBENCH_" in key:
            ev[key.split("LLMDBENCH_")[1].lower()] = value

    # Extract required environment variables with defaults
    is_openshift = ev.get("control_deploy_is_openshift", "0") == "1"
    work_dir = ev.get("control_work_dir", ".")
    current_step = ev.get("current_step", os.path.splitext(os.path.basename(__file__))[0])
    kubectl_cmd = ev.get("control_kcmd", "kubectl")
    dry_run = ev.get("control_dry_run") == '1'
    verbose = ev.get("control_verbose") == '1'
```

yossiovadia: you mean something like this, in functions.py? 

```
    """Parse LLMDBENCH_* environment variables into a dictionary.
    
    Returns:
        dict: Dictionary with lowercase keys (without LLMDBENCH_ prefix) mapping to environment values
    """
    ev = {}
    for key, value in os.environ.items():
        if "LLMDBENCH_" in key:
            ev[key.split("LLMDBENCH_")[1].lower()] = value
    return ev
 ```

maugustosilva: Yeah, something like this, especially because we could then set any defaults (i.e., under `# Extract required environment variables with defaults`) directly there, only refer to `ev[var_name]`).

yossiovadia: Note 1 that it required a little refactoring of all the files created so far, including 04 ( not done by me )

Note 2 -I notice there's an syntax error on line ~130 in 04_ensure_model_namespace_prepared.py It
  has bash-style syntax [[ ]] , i'll also addressed it.

yossiovadia: ok, how about let's finish with all conversion, i'll then refactor all ?

maugustosilva (approved):

PR #280, opened 2025-08-18
Fix improper use of `any`

namasl:
Fix improper type hints using `any` with `Typing.Any`.


achandrasekar (approved):

namasl:
> It might make sense to enable type check and linting for the python code so it is more maintainable in the long run. Wdyt?

Yes, this is absolutely necessary.

PR #318, opened 2025-08-29
[Standup] fix: restore the ability to deploy stacks

maugustosilva:
In order to have a completely functional deployment (with the `gateway` object properly configured), always remove the
`infra-${LLMDBENCH_VLLM_MODELSERVICE_RELEASE}` chart on `teardown`

Some additional fixes, and one improvement: added the ability to have sub-directories on `scenarios` folder


setup/functions.py
```diff
@@ -665,12 +665,16 @@ def add_config(obj_or_filename, num_spaces=0, label=""):
...
+
+    contents = obj_or_filename
+
+    if len(obj_or_filename.split('\n')) == 1 :
+        try:
+            with open(obj_or_filename, 'r') as f:
+                contents = f.read()
+        except FileNotFoundError:
+            # not a file
+            contents = obj_or_filename
```
jgchn, line 677: Nit: can we just `pass` in the except block since it doesn't change the value of `contents`?

PR #360, opened 2025-09-19
Clarify installation steps for config explorer

jgchn:
Clarify steps to installing the UI of config explorer, since it's separate from the lib but requires installing lib first.

### DIFF
--- a/config_explorer/README.md
+++ b/config_explorer/README.md
@@ -49,7 +49,7 @@ Running the Streamlit frontend requires cloning the `llm-d-benchmark` repo. Make
 ```
 git clone https://github.com/llm-d/llm-d-benchmark.git
 pip install -r config_explorer/requirements-streamlit.txt
-streamlit run config_explorer/Home.py
+.venv/bin/streamlit run config_explorer/Home.py
 ```
 
 ### Library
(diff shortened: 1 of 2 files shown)


toslali-ibm (approved):

config_explorer/README.md
```diff
@@ -44,12 +44,12 @@ There are two ways to interact with the Configuration Explorer: via an experimen
...
 A Streamlit frontend is provided to showcase the capabilities of the Configuration Explorer rapidly. Since the core functions are in a module, users may feel free to build their own frontend, such as a CLI, by making use of those functions.
 
-Running the Streamlit frontend requires cloning the `llm-d-benchmark` repo.
+Running the Streamlit frontend requires cloning the `llm-d-benchmark` repo. Make sure you have already installed the required dependencies for the `config_explorer` package following the Installation guide [here](#installation).
 
 ```
 git clone https://github.com/llm-d/llm-d-benchmark.git
 pip install -r config_explorer/requirements-streamlit.txt
-streamlit run config_explorer/Home.py
+.venv/bin/streamlit run config_explorer/Home.py
```
namasl, line 52: If they already activated their virtual environment as described above, then `streamlit run config_explorer/Home.py` would work, no need to add the explicit path.

namasl: If you have an install elsewhere (in your default path), then you activate the virtual environment, doesn't that activation override everything else (like it does for `python`, `pip`...)?

PR #378, opened 2025-09-26
Add Inference Perf configuration to benchmark report

namasl:


namasl:
> What about `stage`?

Oh, right, we need that too. Let's do another PR for that

PR #379, opened 2025-09-26
Add Inference Perf stage number to benchmark report

namasl:


jgchn (commented):
Still dependent on the filename... wondering if we should obtain this info from harness pod in the future

PR #215, opened 2025-08-05
[Setup] bugfix: ensure functionality with Kubernetes/Minikube

maugustosilva:
Additionally, ensure `-n/--dry-run` works for both `standup.sh` and `run.sh`

PR #388, opened 2025-09-30
[Setup] bugfix: avoid error on resource type route in Kubernetes/Minikube

mengmeiye:
This PR fixed the `error: the server doesn't have a resource type "route"` in k8s/minikube when running `standup.sh` at step `10_smoketest`


mengmeiye:
Related PR: https://github.com/llm-d/llm-d-benchmark/pull/215

Issue #386, opened 2025-09-30
LLMDBENCH_HF_TOKEN should be optional


jgchn:
### Component

Setup/Standup

### Describe the bug

LLMDBENCH_HF_TOKEN should be optional for models that do not require it. Fail earlier (before model download step) if token not provided for models that are gated. 

### Steps to reproduce

Trying to deploy a model that doesn't require hf token like `facebook/opt-125m` causes 

```
❌ Environment variable 'LLMDBENCH_HF_TOKEN' is not set.
Error: Process completed with exit code 1.
```

Vezio:
I spoke with @maugustosilva - I have a fix I will put in a PR. Assigned to me.

Issue #290, opened 2025-08-22
Include pylint


pancak3:
### Component

Setup/Teardown

### Desired use case or feature

1. Unify Python code style.
2. Trigger in pre-commit.

### Proposed solution

1. add pylint
2. modify pre-commit config

pancak3:
@maugustosilva Hi Marcio, is there agreement on the standards of code styles?

maugustosilva:
Hi @pancak3. Since we still transitioning from `bash` to `python` (a multi-step process, nowhere near the end) there are not any agreements in place. Proposals are welcome. Thanks.

pancak3:
@maugustosilva, the decision on code styles would be that the earlier, the better. During my contribution to PyScripts, I had to disable my auto-format config, which is annoying. 
I propose to  use the linter of big repos as a template, e.g., https://github.com/pytorch/pytorch/blob/main/.flake8 
This could be a temporary decision, but the final decision needs to be made in a meeting imo.

Vezio:
Please note that:

```

[tool.ruff]
target-version = "py39"
line-length = 119
```

Would need to follow 3.11 or higher version of python do to recent changes.

👍 on the linter @pancak3 are you still ok to take this on?

@maugustosilva thoughts?

maugustosilva:
@pancak3 my apologies for not responding sooner, this is issue fell through the cracks (thanks for the bumping of the issue @Vezio) Yes, let's go with your proposal. I will assign the issue to you then. Tyvm.

pancak3:
Sure, @maugustosilva @Vezio will let you know when the pr is ready

pancak3:
@maugustosilva After further research, we may need to resolve some existing issues before adding it.

We are likely to use ruff as the liner. The configuration of the ruff is better put in [`pyproject.toml`](https://github.com/llm-d/llm-d-benchmark/blob/227d1c83a77973eb0616bd2554b2ddd1ba56afbd/pyproject.toml#L1). However, this file is occupied by a submodule called `config_explorer`. `pyproject.toml` is oddly set to its submodule name in #237 , which could be avoided by using `PYTHONPATH`. I have created #403 to fix this.

Thus, this issue, #290, is blocked by #403.

PR #398, opened 2025-10-01
[Issue-386] Allow Users to Provide HF Credential on As-Needed Basis

Vezio:
# Summary

This PR addresses #386 by checking **_early_** if a model desired to be used is **_gated_** or if it is publicly available. 

This PR adds the ability to _check_ if a model is gated and _validate_ if user has access to that model by checking their HF Token against the permissions of that model. We also include several utility functions and enhancements to various functions.

We preserved the ability to run `env.sh` as well as future proofed our desire to move to `python` by providing both `shell` and `python` implementations of the useful functions enabling this work.

As a misc. item - this also fixes some very trivial `python` standards.

## Key Notes

- Users no longer need to have `LLMDBENCH_HF_TOKEN` if they are deploying `non-gated` models
- Automatic _checking_ is performed to determine if the user **_needs_** to provide a `LLMDBENCH_HF_TOKEN` token AND if they have access to the gated model, even after providing the token - since MANY of them require manual access requests


setup/functions.py
```diff
@@ -1138,20 +1286,102 @@ def check_affinity():
...
+        ) as response:
+            if response.status_code == 200:
+                return True
+            elif response.status_code in (401, 403):
+                return False
+            else:
+                response.raise_for_status()
+    except requests.RequestException as e:
+        announce("❌ ERROR - Request failed:", e)
+        return False
```
jgchn, line 1387: I wonder if there's potential opportunity to exercise the `config_explorer` package here. It includes functions like `get_model_config_from_hf` which checks if model can be accessed. For example, the client code in step 0 looks something like [this](https://github.com/llm-d/llm-d-benchmark/blob/1f44e2273a6116250dec5519102eb5ad67004f50/setup/steps/00_ensure_llm-d-infra.py#L97).

However, if env.sh is invoked before standup.sh, then probably this is not possible?

setup/functions.sh
```diff
@@ -1162,3 +1162,40 @@ function backup_work_dir {
...
+    if [[ $? -ne 0 || -z "${response}" ]]; then
+        return 2
+    fi
+    
+    local gated=$(echo "${response}" | jq -r '.gated // false')
+    if [[ ${gated} == "false" ]]; then
+      return 1
+    else
+      return 0
+    fi
```
jgchn, line 1183: Standard bash pattern is returning 0 for success and 1 for error. Minor nit and not required, but wondering if you want to return 0 for non-gated models and 1 for gated or maybe just add a comment for this function that 1 means non-gated?

setup/steps/04_ensure_model_namespace_prepared.py
```diff
@@ -66,49 +78,79 @@ def add_scc_to_service_account(api: pykube.HTTPClient, scc_name: str, service_ac
...
+        api=api,
+        namespace_name=ev["vllm_common_namespace"],
+        dry_run=ev["control_dry_run"],
+    )
 
     if ev["hf_token"]:
-        announce(f'🔑 Creating or updating secret "{ev["vllm_common_hf_token_name"]}"...')
+        announce(
+            f'🔑 Creating or updating secret "{ev["vllm_common_hf_token_name"]}"...'
+        )
```
jgchn, line 113: Do we need to update ev['hf_token']? If `LLMDBENCH_HF_TOKEN=""` or `LLMDBENCH_HF_TOKEN=`, I think this will just be an empty string for `ev['hf_token']`. Empty strings are invalid tokens when invoking hf api. Do we need to convert `ev['hf_token'] == ""` into `None`?

setup/env.sh
```diff
@@ -475,14 +475,29 @@ if [[ $LLMDBENCH_CONTROL_CALLER == "run.sh" ]]; then
...
+export HF_TOKEN=${HF_TOKEN:-$LLMDBENCH_HF_TOKEN}
+
+if ! echo ${LLMDBENCH_CONTROL_CALLER} | grep -iq "teardown"; then
+  if is_hf_model_gated "${LLMDBENCH_DEPLOY_MODEL_LIST}"; then
+    if [[ -z ${HF_TOKEN} ]]; then
+      announce "❌ Hugging Face Token is empty but attempted to use gated model \"${LLMDBENCH_DEPLOY_MODEL_LIST}\""
+      exit 1
+    fi
+
+    if user_has_hf_model_access "${LLMDBENCH_DEPLOY_MODEL_LIST}" "${HF_TOKEN}"; then
```
jgchn, line 487: `LLMDBENCH_DEPLOY_MODEL_LIST` could be a list of models, comma separated. Need for loop here?

setup/steps/04_ensure_model_namespace_prepared.py
Vezio: No. Since `ev["hf_token"]` will not be defined. It will not enter this loop. If `ev["hf_token"]` that means the user did not provide any token and intends to use a non-gated model. Based on my implementation, if the user _tried_ to access a gated model, the code will ensure they're notified earlier.

I have ensured this is not an issue and are no empty strings :)

jgchn: Nice!

setup/functions.py
Vezio: This is a conversation I had with @maugustosilva - there is an **obvious** API that huggingface provides that renders this curl as obsolete - but during that discussion we opted to NOT include it as a package.

But now I see it is _here_ so yes we can leverage that. But `shell` would not really leverage that `python` package at all...

Vezio: @jgchn I'll expand on that package -- looks like a pretty good spot. Still will leave the bash impl. here until we are totally done away with it. Thanks for this!!!

maugustosilva:
@namasl I am perfectly ok with a PR for just style. We actually had this as an issue for quite a while #290

maugustosilva (dismissed):

maugustosilva (approved):

PR #407, opened 2025-10-03
[Issue 401] Update Dependency Checker and Add Versioning

### LINKED ISSUE #401: Requirements and Dependencies are not Checked Appropriately
### Component

Setup/Standup

### Describe the bug

There is no accurate source of truth for the dependencies needed up front.

Currently, the code base does not appropriately _check_ if all dependencies are installed that are required for running the `llm-d-benchmark` suite. 

Things like `huggingface_hub` and `python version` (for example, now we require 3.11 and up - this may break existing users environments and would be a _nice to have up front_)  and others, need to be checked and/or appropriately installed in the `install_deps.sh` script as we have done so before.

It appears that our `install_deps.sh` has lagged behind with the rapid amount of progress done, we should bring that up to speed and accurately document the dependencies and their versions needed run the `llm-d-benchmark` suite. 

There have been several reports of users running into issues recently.

### Steps to reproduce

Running `standup.sh`, for example,  or simple scripts that require dependencies.

Vezio:
# Summary
Updates dependencies required for successful installation - this will eventually _go away_ after our Python full conversion - but to avoid _breaking_ users during the intermediate time period, let's _prevent_ new packages and versions from being scattered through our evolving code base.

- Enhances dependency checker and installer 
- Adds ability to install version specific pip packages
- Checks if python3.11 >= is available on system as this is now a requirement.

Should resolve #401

### DIFF
--- a/setup/install_deps.sh
+++ b/setup/install_deps.sh
@@ -159,7 +159,7 @@ if ! command -v pip3 &> /dev/null; then
     echo "pip3 installed successfully."
 fi
 
-python_deps="kubernetes pykube-ng kubernetes-asyncio GitPython requests PyYAML Jinja2 huggingface_hub==0.34.4 transformers==4.55.4"
+python_deps="kubernetes pykube-ng kubernetes-asyncio GitPython requests PyYAML Jinja2 requests huggingface_hub==0.34.4 transformers==4.55.4"
 
 for dep in $python_deps; do
     pkg_name=$(echo "${dep}" | cut -d= -f1)
--- a/setup/install_deps.sh
+++ b/setup/install_deps.sh
@@ -115,6 +115,30 @@ for tool in $tools; do
     echo "---------------------------"
 done
 
+# 
+# Check minimum Python version (3.11+) based on new requirements
+# 
+python_present=""
+for pybin in python3 python3.{13..11}; do
+    if command -v ${pybin} &>/dev/null; then
+        ver=$(${pybin} -c 'import sys; print(".".join(map(str, sys.version_info[:2])))')
+        major=$(echo ${ver} | cut -d. -f1)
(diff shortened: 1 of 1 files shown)


setup/install_deps.sh
```diff
@@ -135,21 +159,32 @@ if ! command -v pip3 &> /dev/null; then
     echo "pip3 installed successfully."
 fi
 
-python_deps="kubernetes pykube-ng kubernetes-asyncio GitPython requests PyYAML Jinja2"
+python_deps="kubernetes pykube-ng kubernetes-asyncio GitPython requests PyYAML Jinja2 requests huggingface_hub==0.34.4 transformers==4.55.4"
```
jgchn, line 162: Should we have a general requirements.txt file in the project root and pin versions for each dependency?

Vezio: This question could have a much longer answer, but I’ll try to keep it brief and share my perspective:

As mentioned in the summary, this isn’t meant to be a permanent solution. The goal here is simply to get our current users unblocked by working within the design we already have.

The long-term solution will come as part of the python refactor, which should address many of the pain points we’re dealing with now. Since we’re not at that stage yet, I put together this minimal fix so users aren’t stuck in the meantime.

Right now, besides the bash scripts we’ve converted to Python, we also have a Python package in the codebase (config_explorer). It runs somewhat independently with its own requirements.txt, and this PR was needed to make sure it works correctly for users.

Neither the current “one-size-fits-all” requirements.txt nor this stopgap solution is intended to survive the refactor—this is just about keeping things working until the more complete solution is in place.

maugustosilva: This is one step in the right direction. I will merge as-is and we will keep refining it.

maugustosilva (approved):

PR #410, opened 2025-10-04
Refactor project structure to python proj

pancak3:


pancak3:
> @pancak3 I'd propose the following integration path:
> 
> * Open a PR with changes for integrating the linter to automation (but not directly using it, yet)
> * Open a PR with changes for the changes made _by_ the linter (then enable the linter automation)
>   
>   * For this one - I'd prefer if we made the changes in _chunks_ so we can transparently view what was changed
> * Open a PR with changes for the Python Package restructure
> 
> Thank you for such great efforts!

That sounds pretty reasonable to me! I will check out the first todo item within the week and others after. Got some ddls to catch this month. Thanks!

Vezio:
@pancak3 awesome thanks! I sense there will be a LOT of contributions this week and a bit of refactoring, so it will probably be good timing for your research you have going on so the code takes a breather 😄

Issue #404, opened 2025-10-03
Incorrect stack endpoint detection


poussa:
### Component

Run

### Describe the bug

I have a llm-d stack deployed via `llm-d/guides/llm-d/guides/inference-scheduling` with the command:

```RELEASE_NAME_POSTFIX=xpu helmfile apply -n llm-d -e xpu```

I get the gateway:

```console
k get gateway
NAME                          CLASS   ADDRESS                                                     PROGRAMMED   AGE
infra-xpu-inference-gateway   istio   infra-xpu-inference-gateway-istio.llm-d.svc.cluster.local   True         4h5m
```
Run.sh script incorrectly detects the endpoint to be:
```console
 ℹ️ Stack Endpoint URL detected is "http://infra-xpu-inference-gateway.llm-d.svc.cluster.local:80/meta-llama-llama-3-2-1b-instruct"
```
NOTE: it is missing the `-istio`

The quick fix below works for me but don't know what it may break.

```diff
@@ -242,7 +242,8 @@ for method in ${LLMDBENCH_DEPLOY_METHODS//,/ }; do
       if [[ $LLMDBENCH_CONTROL_ENVIRONMENT_TYPE_MODELSERVICE_ACTIVE -eq 1 ]]; then
         export LLMDBENCH_CONTROL_ENV_VAR_LIST_TO_POD="LLMDBENCH_BASE64_CONTEXT_CONTENTS|^LLMDBENCH_VLLM_COMMON|^LLMDBENCH_VLLM_MODELSERVICE|^LLMDBENCH_DEPLOY|^LLMDBENCH_VLLM_INFRA|^LLMDBENCH_VLLM_GAIE|^LLMDBENCH_LLMD_IMAGE"
         export LLMDBENCH_HARNESS_STACK_TYPE=llm-d
-        export LLMDBENCH_HARNESS_STACK_ENDPOINT_NAME=$(${LLMDBENCH_CONTROL_KCMD} --namespace "$LLMDBENCH_VLLM_COMMON_NAMESPACE" get gateway --no-headers | grep ^infra-${LLMDBENCH_VLLM_MODELSERVICE_RELEASE}-inference-gateway | awk '{print $1}')
[...]

PR #419, opened 2025-10-07
[Standup] Add labels `stood-up-by`, `stood-up-from` and `stood-up-via` to objects stood up by llm-d-benchmark

maugustosilva:
We then use these labels in `run.sh` to determine the model URL when `standup.sh` was executed.

When a suitable URL is not found, an informative message is displayed, which should help prevent the reocurrence of #404.


Vezio (approved):
lgtm

PR #421, opened 2025-10-07
Adding new function for kv_cache_per_token to config_explorer

Edwinhr716:


jgchn (commented):
LGTM, could you please add a unit test? Also I think `KVCacheDetail` could probably be used directly without having to expose each field as a function. Thoughts on this? @Edwinhr716 would like to understand your use case also to see how we can make this lib more usable.

Edwinhr716:
Mmm yeah I was considering that as well, in general would prefer not to create a KVCacheDetail and abstract that away as a user of the python module. That way all I have to do is call the appropriate functions instead of having to instantiate a class. I saw that you do a similar thing for kv_cache_req, and thought to add it for kv_cache_per_token as well.

Edwinhr716:
As for my use case, I wish to collect data relevant for each model that I'm interested in, such as model_len, attention heads, etc to better compare models in terms of hardware required to run them.

I had already created my own, but I saw that a library already existed to cover my use case, so decided to contribute instead.

jgchn:
> That way all I have to do is call the appropriate functions instead of having to instantiate a class.

Makes sense. Originally we just have `kv_cache_req` as a function, but realize it's important to expose the internal mechanics of how that number is derived, hence the creation of KVCacheDetail with all the relevant parts. 

> I wish to collect data relevant for each model that I'm interested in, such as model_len, attention heads, etc to better compare models in terms of hardware required

It sounds like you could still benefit from KVCacheDetail? It collects info you're interested in

PR #423, opened 2025-10-08
Tekton Orchestrated Benchmarking POC

kalantar:
See tekton_poc/README.md


charts/harness/Chart.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v2
```
sriumcp, line 1: Is this chart needed?

Can we simply use the inference-perf container as part of the StepAction, and configure it so that the step actually directly sends the Inference request?

This approach eliminates the chart, along with the need for a separate harness pod (the Tekton step will run as part of some pod anyway).

sriumcp (changes requested):

sriumcp (changes requested):

jgchn (commented):
Tested and looks good! I was able to execute three experiments in parallel asynchronously, and saw the results uploaded to IBM COS. Lets get this merged!

kalantar:
This work has moved to https://github.com/inference-sim/tektonc-data-collection

PR #452, opened 2025-10-17
Add total kv blocks func

### LINKED ISSUE #454: Config_explorer/capacity_planner: need to compute total kv blocks for a given model/hw/parallelism
### Component

Config_explorer/capacity_planner

As a user of config explorer capacity planner, I would like if it exposes a functionality to compute total kv blocks for a given model/hw/parallelism. 

Also, the current `per_token_memory_bytes` formula is not correct (as emprically validated - see attached pictures)

Image

Image

### Proposed solution

- Introduce `total_kv_cache_blocks` func into capacity planner
- Fix incorrect formula of `per_token_memory_bytes`

### Alternatives

_No response_

_No response_

toslali-ibm:
Resolves #454 
- Compute the total number of KV cache blocks available for a model on a GPU.
- Fix incorrect formula of `per_token_memory_bytes`


config_explorer/src/config_explorer/capacity_planner.py
```diff
@@ -396,6 +396,30 @@ def kv_cache_req(model_info: ModelInfo,
...
+    per_token_memory = kv_cache_detail.per_token_memory
+    per_block_memory = per_token_memory * block_size
+
+    # Compute allocatable KV cache memory
+    kv_cache_allocatable = allocatable_kv_cache_memory(
+        model_info, model_config, gpu_memory, gpu_mem_util
+    )
+
+    # Compute total KV cache blocks
+    total_kv_blocks = int((kv_cache_allocatable * (1024 ** 3)) // per_block_memory)
```
jgchn:
```suggestion
    total_kv_blocks = int(bytes_to_gib(kv_cache_allocatable) // per_block_memory)
```

Instead of converting to int, should it be `math.floor`?

config_explorer/src/config_explorer/capacity_planner.py
```diff
@@ -396,6 +396,30 @@ def kv_cache_req(model_info: ModelInfo,
 
     return KVCacheDetail(model_info, model_config, context_len, batch_size).kv_cache_size_gb
 
+def total_kv_cache_blocks(model_info: ModelInfo,
+                    model_config: AutoConfig,
+                    context_len: int,
+                    gpu_memory: int,
+                    gpu_mem_util: float=0.9,
+                    batch_size: int = 1,
+                    block_size: int = 16,
```
jgchn: I'm also wondering if we want to do input checking? vLLM only accepts certain values for block size. Future enhancement, perhaps, but would be good to make a note of it in docstring

config_explorer/src/config_explorer/capacity_planner.py
toslali-ibm: Here, I am trying the opposite actually -> gib_to_bytes. I added this function now. 

Not sure if it should be floor() or int()

jgchn: It looks like `//` is floor division in Python https://docs.python.org/2/reference/expressions.html#binary-arithmetic-operations

jgchn (approved):

PR #457, opened 2025-10-18
Add configuration exploration capabilities to config_explorer library

namasl:
This PR adds functions for importing a set of benchmark report files to a Pandas DataFrame, and grouping/filtering/plotting the results. This allows one to gain insights from a large collection of benchmark data.

A demonstration of the library's usage can be seen in the Jupyter notebook [analysis.ipynb](https://github.com/namasl/llm-d-benchmark/blob/ac43e268e362262390fba903fc77b15c6755f098/analysis/analysis.ipynb), which replaces the standalone analysis notebooks. With most code hidden in the library, the notebook becomes significantly slimmer despite now being much more flexible.

A follow-on PR from @jgchn will demonstrate the library's capabilities through an interactive Streamlit front-end. This example will more closely represent how an end user would use benchmark data to get direct feedback on how to best configure their cluster for their use case (rather than the notebook which is developer/researcher focused).


config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1234 @@
...
+    """
+    report = convert.import_benchmark_report(br_file)
+
+    # Get plugin parameters
+    prefix_cache_scorer_block_size = None
+    prefix_cache_scorer_lur_capacity_per_server = None
+    prefix_cache_scorer_max_blocks_to_match = None
+    prefix_cache_scorer_tracking = False
+    if report.scenario.platform.metadata:
+        for plugin in report.scenario.platform.metadata['inferenceScheduler']['plugins']:
```
jgchn: Should you first check if `inferenceScheduler` is a key in `.scenario.platform.metadata`? I guess the PD scenario doesn't have `.report.scenario.platform.metadata` section, so we're under the assumption that this script only works for the data on Drive, right?

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1234 @@
...
+
+    def __init__(self, col: str, value: float):
+        if col not in COLUMNS:
+            raise ValueError(f'Column does not exist: {col}')
+        if COLUMNS[col].dtype != 'float':
+            raise TypeError(f'Column must have float datatype: {col}')
+        if COLUMNS[col].pref == Pref.NEUTRAL:
+            raise Exception(f'Column must have a preferred direction: {col}')
+        self.col = col
+        self.value = value
```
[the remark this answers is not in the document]
namasl: Thanks, updated to use this.

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1234 @@
...
+        'Prefix_Cache_Scorer_Weight': prefix_cache_scorer_weight,
+        'Prefix_Cache_Scorer_Block_Size': prefix_cache_scorer_block_size,
+        'Prefix_Cache_Scorer_LRU_Capacity_Per_Server': prefix_cache_scorer_lur_capacity_per_server,
+        'Prefix_Cache_Scorer_Max_Blocks_To_Match': prefix_cache_scorer_max_blocks_to_match,
+        'Prefix_Cache_Scorer_Tracking': prefix_cache_scorer_tracking,
+        # Workload
+        'Workload_Generator': report.scenario.load.name,
+        'ISL': int(round(report.metrics.requests.input_length.mean)),
+        'OSL': int(round(report.metrics.requests.output_length.mean)),
+        'ISL_500': int(round(report.metrics.requests.input_length.mean/500)*500+250),
```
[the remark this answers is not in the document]
namasl: This (likewise with `OSL_500`) are the sequence lengths rounded to the nearest 500. The addition of `250` is so that the value displayed is the central value of the bin. I would like to replace this with a more flexible method of binning, but went with something simpler to start with.

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+    get_nested(report.scenario.platform.metadata, ['inferenceScheduler', 'plugins']):
+        for plugin in report.scenario.platform.metadata['inferenceScheduler']['plugins']:
+            if plugin.get('type') == 'prefix-cache-scorer':
+                if 'parameters' not in plugin:
+                    continue
+                prefix_cache_scorer_block_size = plugin['parameters'].get('blockSize', 16)
+                prefix_cache_scorer_lur_capacity_per_server = plugin['parameters'].get('lruCapacityPerServer', 31250)
+                prefix_cache_scorer_max_blocks_to_match = plugin['parameters'].get('maxPrefixBlocksToMatch', 256)
+                # If mode is 'cache_tracking', then precise prefix scoring is used
+                prefix_cache_scorer_tracking = plugin['parameters'].get('mode', '') == 'cache_tracking'
```
jgchn:
```suggestion
                if 'parameters' in plugin:
                    params = plugin['parameters']
                    prefix_cache_scorer_block_size = params.get('blockSize', 16)
                    prefix_cache_scorer_lur_capacity_per_server = params.get('lruCapacityPerServer', 31250)
                    prefix_cache_scorer_max_blocks_to_match = params.get('maxPrefixBlocksToMatch', 256)
                    # If mode is 'cache_tracking', then precise prefix scoring is used
                    prefix_cache_scorer_tracking = params.get('mode', '') == 'cache_tracking'
```

Why not capture 'mode' as a column?

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+        concurrency = None
+
+    max_qps = None
+    if report.scenario.load.name == schema.WorkloadGenerator.INFERENCE_PERF:
+        # Workload generator stage
+        stage = report.scenario.load.metadata.get('stage', -1)
+        if stage >= 0:
+            stage_list = get_nested(report.scenario.load.args, ['load', 'stages'])
+            if stage_list and len(stage_list) > stage:
+                 max_qps = stage_list[stage].get('rate')
```
jgchn: A bit confused here. Is `.scenario.load.metadta.stage` is the number of stages? The why would `len(stage_list)` be greater than `stage`?

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+
+    rp = _get_replicas_and_parallelism(report)
+    if rp['is_pd']:
+        num_gpus = 0
+        if rp['p_replicas']:
+            num_gpus += rp['p_tp']*rp['p_replicas']
+        if rp['d_replicas']:
+            num_gpus += rp['d_tp']*rp['d_replicas']
+    else:
+        num_gpus = rp['tp']*rp['replicas']
```
jgchn:
```suggestion
    num_gpus = 0
    if rp['is_pd']:
        if rp['p_replicas']:
            num_gpus += mul(rp['p_tp'], rp['p_replicas'])
        if rp['d_replicas']:
            num_gpus += mul(rp['d_tp'], rp['d_replicas'])
    else:
        num_gpus = mul(rp['tp'], rp['replicas'])
```

jgchn (commented):
Could you also add a section to `config_explorer/README.md` that demonstrates the usage of the `explorer` lib? Or link to the analysis.ipynb?

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+        dtype='float',
+        label='Queue',
+    ),
+    'Prefix_Cache_Scorer_Weight': ColumnProperties(
+        dtype='float',
+        label='Prefix Cache',
+    ),
+    'Prefix_Cache_Scorer_Block_Size': ColumnProperties(
+        dtype='int',
+        label='Block Size',
```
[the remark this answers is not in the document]
namasl: These are labels used for plots and tables. In some cases it is the same as the column name, in other cases like `P90_TTFT_ms` it becomes the cleaner label `Time to First Token P90` which is used in conjunction with the `units` field to independently describe the units as needed.

config_explorer/src/config_explorer/explorer.py
namasl: This was just added in case the data in the report isn't formatted as expected, though there shouldn't be a reason to expect this to happen. I can remove these checks.

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+            rp['p_dp'] = accel.parallelism.dp
+            rp['p_pp'] = accel.parallelism.pp
+            rp['p_ep'] = accel.parallelism.ep
+        if report.scenario.host.type[ii] is schema.HostType.DECODE and not rp['d_tp']:
+            rp['d_tp'] = accel.parallelism.tp
+            rp['d_dp'] = accel.parallelism.dp
+            rp['d_pp'] = accel.parallelism.pp
+            rp['d_ep'] = accel.parallelism.ep
+        if rp['p_tp'] and rp['d_tp']:
+            break
```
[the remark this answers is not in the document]
namasl: We're just trying to populate the parallelism fields for P and D. Once we've got those details we're done.

config_explorer/src/config_explorer/explorer.py
```diff
@@ -0,0 +1,1237 @@
...
+        if rp['d_replicas']:
+            num_gpus += rp['d_tp']*rp['d_replicas']
+    else:
+        num_gpus = rp['tp']*rp['replicas']
+
+    thpt_per_gpu = report.metrics.throughput.output_tokens_per_sec/num_gpus
+    if concurrency:
+        thpt_per_user = report.metrics.throughput.output_tokens_per_sec/concurrency
+    else:
+        thpt_per_user = None
```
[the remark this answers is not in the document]
namasl: No, `num_gpus` is always non-zero, and we have the `if` statement for concurrency.

config_explorer/src/config_explorer/explorer.py
```diff
@@ -838,12 +838,12 @@ def _get_replicas_and_parallelism(report: schema.BenchmarkReport) -> dict[str, i
     # We have a P/D setup
     rp['is_pd'] = True
     for ii, accel in enumerate(report.scenario.host.accelerator):
-        if report.scenario.host.type[ii] is schema.HostType.PREFILL and not rp['p_tp']:
+        if report.scenario.host.type[ii] is schema.HostType.PREFILL and rp['p_tp'] is None:
```
jgchn: To this point `rp['p_tp']` has not been modified. Is `rp['p_tp'] is None:` necessary?

config_explorer/src/config_explorer/plotting.py
```diff
@@ -0,0 +1,341 @@
...
+        # Make a DataFrame for specific configuration
+        conf_df = runs_df
+        label = ''
+        for jj, val in enumerate(conf):
+            conf_df = conf_df[(conf_df[config_keys[jj]] == val)].sort_values(by=col_x)
+            if config_keys[jj] == col_seg_by:
+                continue
+            label += f'{COLUMNS[config_keys[jj]].label}={val}, '
+        # Remove trailing ", "
+        label = label.rsplit(', ', 1)[0]
```
jgchn: How about 

```py
labels = []
for j:
    labels.append(f-string)
label = labels.join(" ,")
```

config_explorer/src/config_explorer/explorer.py
namasl: We are looping through all accelerator setups until we find a prefill setup, then grabbing its details. Likewise with decode. We only want to fill in each of these once, but there is no guarantee when we will get an accelerator setup for prefill or decode.

jgchn (approved):

PR #465, opened 2025-10-21
tektonc poc

sriumcp:
A quick PoC to illustrate composition, parallelism, retries (on failure), and other good stuff that using Tekton enables for us in the context of llm-d-benchmarking.


kalantar:
This work has moved to https://github.com/inference-sim/tektonc-data-collection.

PR #470, opened 2025-10-23
Return Figure object from plot functions

namasl:
- Return `matplotlib.pyplot.Figure` objects from plotting functions (rather than immediately executing `show()`)
- Properly scale Z label offsets on log scaled plots
- PEP8 formatting


config_explorer/src/config_explorer/plotting.py
```diff
@@ -317,6 +347,10 @@ def plot_pareto_tradeoff(
     else:
         plot_func = plt.plot
 
+    global fignum
```
jgchn: Why does this have to be global?

jgchn (approved):

PR #472, opened 2025-10-24
Add function to get scenario counts

namasl:


jgchn (commented):
I meant, as part of the scenario df that is returned in `get_scenario()`, include a count column. Sorry if that wasn't clear. But the table that is printed in `print_scenarios` is nice, and I was hoping to have get_scenarios() or another function to match that.

jgchn (approved):

PR #473, opened 2025-10-25
Config Exploration Visualizer enhancements

jgchn:
- Adds advance SLO filtering


namasl:
The SLO input fields are restricted to `int`, can we make them `float`?

namasl:
Performance tradeoff comparison can have a drop-down for Z.

Pareto plot should also get configurable X/Y

PR #484, opened 2025-10-29
Add ability to set bounds in config explorer library

namasl:
Currently, scenarios (a set of columns in the dataset that will be held constant) must be a single fixed value. This PR allows for scenario columns to be defined by bounds.

For example, the output sequence length may be part of the columns defining a scenario to be explored. Rather than restricting to benchmark results that are an exact value of 1000, with this PR we may now select any results with OSL between 800 and 1200.

This is best demonstrated in the inference scheduler section of `analysis.ipynb`, where scenarios available are printed with available ranges, and from histograms of available values for certain columns, the user can set bounds to select a group of data.

### DIFF
--- a/config_explorer/src/config_explorer/plotting.py
+++ b/config_explorer/src/config_explorer/plotting.py
@@ -118,7 +118,7 @@ def _make_title(scenario: dict[str, Any]) -> str:
                 continue
             value = scenario[col_bound]
             val_bounds.append(f'{COLUMN_BOUND_STR[bound_type]}{value}')
-        title += f'{COLUMNS[col].label}: {' '.join(val_bounds)}  '
+        title += f'{COLUMNS[col].label}: {" ".join(val_bounds)}  '
 
     return title.strip()
 


config_explorer/src/config_explorer/explorer.py
```diff
@@ -1122,36 +1173,89 @@ def add_benchmark_report_to_df(
...
+                    s_dict['__ge__' + col_num] = val_min
+                    s_dict['__le__' + col_num] = val_max
+        else:
+            for ii, col in enumerate(scenario_columns):
+                s_dict[col] = s_tuple[ii]
+        if all_na:
+            # Do not add this scenario
+            all_na = False
+        else:
+            scenarios.append(s_dict)
```
jgchn:
```suggestion
        if not all_na:
            # Add scenario only if column is non empty
            scenarios.append(s_dict)
        all_na = False
         
```

config_explorer/src/config_explorer/explorer.py
```diff
@@ -55,6 +55,42 @@
     from config_explorer import schema
 
 
+# Length of column bound prefix
+BOUND_PREFIX_LEN = 6
```
jgchn: Maybe move all these constants to one file.

config_explorer/pages/2_Sweep_Visualizer.py
```diff
@@ -160,13 +160,13 @@ def add_metric_dialog():
 
     # Remove curr metrics from all performance metrics
 
-    all_metrics =  dict(xp.PERFORMANCE_METRIC_COLUMNS)
+    all_metrics =  dict(xp.METRIC_COLUMNS)
```
jgchn: It should be `METRICS_COLUMNS`

PR #490, opened 2025-10-31
Add more GuideLLM parameters to DataFrame

namasl:


config_explorer/src/config_explorer/explorer.py
```diff
@@ -989,41 +1008,55 @@ def add_benchmark_report_to_df(
...
+    elif report.scenario.load.name == schema.WorkloadGenerator.GUIDELLM:
+        # Workload generator stage
+        # If stage metadata is missing, this benchmark report is from an older
+        # version of convert.py that only took stage 0 results.
+        stage = report.scenario.load.metadata.get('stage', 0)
+        
+        if 'rate' in args:
+            max_qps = args['rate'][stage]
+        concurrenies = get_nested(args, ['profile', 'measured_concurrencies'])
+        if concurrenies:
```
jgchn: Spelling
```suggestion
        concurrencies = get_nested(args, ['profile', 'measured_concurrencies'])
        if concurrencies:
```

config_explorer/src/config_explorer/explorer.py
```diff
@@ -989,41 +1008,55 @@ def add_benchmark_report_to_df(
...
+            concurrency = concurrenies[stage]
+        data_list = args.get('data')
+        if data_list:
+            data = yaml.safe_load(data_list[0])
+            assert isinstance(data, dict)
```
jgchn: How using try-except here instead of `assert`? Assert is usually for testing 

```yaml
data = {}
try: data = yaml.safe_load(data_list[0])
except: pass 
```

namasl: This isn't needed, I just had that in there to make the type checker happy. I'll delete it.

PR #502, opened 2025-11-10
always copy the context.ctx file

effi-ofer:
The context.ctx file is copied only if it does not exist.  But it could be stale.  Why not always copy it?

### DIFF
--- a/setup/env.sh
+++ b/setup/env.sh
@@ -437,55 +437,53 @@ backup_work_dir
 
 prepare_work_dir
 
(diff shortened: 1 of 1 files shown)


effi-ofer:
Both my team mate and I independently encountered a stale context file yesterday.  So it's a real problem.  In my case it happened because I switched In my case it happened because I switched LLMDBENCH_CONTROL_WORK_DIR to a directory that was previously used as the work dir and contained a stale context file, so you could argue it was my fault.  But I think the change avoids the problem all together.

PR #560, opened 2025-12-09
delete app=llm-d-benchmark-harness pods

effi-ofer:
At the end of a run, llm-d-benchmark deletes pods with label app=llmdbench-harness-launcher but does not delete pods with label app=llm-d-benchmark-harness such as access-to-harness-data-*.  These pods need to be deleted, or else we fail the collection of run results at the very end of a run if there are old runs that used different harness.  

Perhaps the right long term approach is use the llmdbench-harness-launcher for the harness data pod?  I wasn't sure, so in this PR is simply added a delete statement.

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -504,6 +504,9 @@ function deploy_harness_config {
         llmdbench_execute_cmd "${LLMDBENCH_CONTROL_KCMD} --namespace ${LLMDBENCH_HARNESS_NAMESPACE} delete pod -l app=${LLMDBENCH_HARNESS_POD_LABEL}" \
             ${LLMDBENCH_CONTROL_DRY_RUN} \
             ${LLMDBENCH_CONTROL_VERBOSE}
+        llmdbench_execute_cmd "${LLMDBENCH_CONTROL_KCMD} --namespace ${LLMDBENCH_HARNESS_NAMESPACE} delete pod -l app=llm-d-benchmark-harness" \
+            ${LLMDBENCH_CONTROL_DRY_RUN} \
+            ${LLMDBENCH_CONTROL_VERBOSE}
         announce "✅ Pods with label \"app=${LLMDBENCH_HARNESS_POD_LABEL}\" for model \"$model\" deleted"
     elif [[ $LLMDBENCH_HARNESS_WAIT_TIMEOUT -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


namasl:
@effi-ofer I'm a bit confused why you are having issues. When are you failing to collect data, and are you seeing some sort of error? What type of experiment?

We use the app label `llm-d-benchmark-harness` for the model `downloader` pod and the `access-to-harness-data-PVC_NAME` pod. The app label given in the `LLMDBENCH_HARNESS_POD_LABEL` envar is used for the harness pod (defaults to `llmdbench-harness-launcher`).

During an experiment, [`deploy_harness_config`](https://github.com/llm-d/llm-d-benchmark/blob/6754f610c6f4140197b6aef28a0effb699a3eaab/setup/functions.sh#L444) does the following in the usual case (not in debug mode, finite timeout):
- Create a harness pod (with app label given in `LLMDBENCH_HARNESS_POD_LABEL`).
- Wait for that pod to complete.
- Copy results and `rsync` analysis to the local machine from the PVC via the `access-to-harness-data-PVC_NAME` pod.
- Delete pods with app label given in `LLMDBENCH_HARNESS_POD_LABEL` (just harness pods).

If the `access-to-harness-data-PVC_NAME` pod gets deleted I don't see how it would recover, and I would expect if an experiment includes several runs, the next launch of a harness pod will fail to collect data.

We could do away with the `access-to-harness-data-PVC_NAME` pod entirely and just copy data from the harness pod directly. I'm guessing the access pod which lingers around is meant as a convenience if the user wants to connect to it. @maugustosilva is this right?

effi-ofer:
If there are multiple access-to-harness-data-PVC_NAME existing during the data collection part of a run, then we get this error:
```
==> Tue Dec  9 10:09:28 IST 2025 - ./run.sh - 🏗️ Collecting results for pods with label "app=llmdbench-harness-launcher"...
ERROR while executing command "oc --kubeconfig /home/effi/llm-d-benchmark/runs/environment/context.ctx --namespace effi cp --retries=5                 access-to-harness-data-effi-vllm-cache42
access-to-harness-data-effi-vllm-cache5:/requests/inference-perf_1765267126_vllm-lmcache-8b-instruct_1 /home/effi/llm-d-benchmark/runs/results/inference-perf_1765267126_vllm-lmcache-8b-instruct_1"
```
In this run, we ended up with the following two access-to-harness-data-PVC_NAME files:

1. access-to-harness-data-effi-vllm-cache5 (from a prior run)
2. access-to-harness-data-effi-vllm-cache42 (from the current run)

We could fix the cp statement, but why not clean up at the start of the run instead?

PR #566, opened 2025-12-11
Benchmark runner for an existing stack

dmitripikus:
This is a simplified version based on `run.sh` for an existing (pre-deployed) llm-d stack.
It does not setup the harness namespace (use `prepare.sh` for this).
It uses a single yaml configuration file. There is no discovery for configuration parameters.
A configuration file could be prepared using `prepare.sh` script (using `run.sh` command line arguments and environment).
The user can customize the configuration file as needed (there is an example of config.yaml). The user can add custom environment variables as needed.

### DIFF
--- a/existing_stack/run_only.sh
+++ b/existing_stack/run_only.sh
@@ -229,9 +229,7 @@ if [[ "$harness_parallelism" != "1" ]]; then
 fi  
 #@TODO harness_parallelism=1 only is supported for now!!!
 
-
-
-_harness_pod_name=$(sanitize_pod_name "llmdbench-${harness_name}-launcher")
+_harness_pod_name=$(sanitize_pod_name "${HARNESS_POD_LABEL}")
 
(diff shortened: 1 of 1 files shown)


dmitripikus:
Hi Marcio @maugustosilva , thank you for the comments!
1. We fully agree that it doesn't make sense to keep two function.sh: 'existing_stack/functions.sh' and 'setup/functions.sh'. 
We thought to add functions from 'existing_stack/functions.sh' to 'existing_stack/run_only.sh' (and have a single file for existing stack benchmark run):
- 'create_harness_pod()' in 'setup/functions.sh' and 'start_harness_pod()' in 'existing_stack/functions.sh' pass different number of env vars
- there is only a small overlap between other functions in these two files
Does this seem reasonable?

2. We'll definitely work on re-writing run shell script to Python

deanlorenz:
- This is used by llm-d/llm-d#559 (currently referencing the fork). 
- Should add a `well-lit` tag to be used for reference from the `llm-d/guides`.

PR #567, opened 2025-12-11
[Standup] Compatibility with the new `istio` (1.28.1)

maugustosilva:


setup/steps/08_deploy_gaie.py
```diff
@@ -150,7 +150,7 @@ def main():
 inferencePool:
   targetPortNumber: {ev['vllm_common_inference_port']}
   modelServerType: vllm
-  apiVersion: "inference.networking.x-k8s.io/v1alpha2"
+  apiVersion: "inference.networking.k8s.io/v1"
```
kalantar: Is it premature to move API version versus making to configurable?  I notice that several llm-d guides still use the older version.

maugustosilva: @kalantar I don't think it is premature, for two reasons: 1) `istio` 1.28.1 is the actual first mainline version to support `llm-d` in its entirety and 2) there is a PR updating all guides at https://github.com/llm-d/llm-d/pull/513

PR #583, opened 2025-12-23
GPU recommender using llm-optimizer's roofline analysis tool

jgchn:
An initial GPU recommender tool using [llm-optimizer](https://github.com/bentoml/llm-optimizer)'s roofline analysis. In the future, use a more accurate model. 

Screenshot 2025-12-23 at 14 02 58

Screenshot 2025-12-23 at 14 03 10

### DIFF
--- a/config_explorer/src/config_explorer/recommender/recommender.py
+++ b/config_explorer/src/config_explorer/recommender/recommender.py
@@ -20,6 +20,7 @@ class GPURecommender:
         output_len: int,
         max_gpus: int = 1,
         max_gpus_per_type: Optional[Dict[str, int]] = None,
+        gpu_list: Optional[list] = None,
 
         # Performance constraints
         max_ttft: Optional[float] = None,
@@ -37,6 +38,7 @@ class GPURecommender:
             max_gpus_per_type: Optional dict mapping GPU names to their specific max_gpus limit.
                               Example: {"H100": 8, "A100": 4, "L40": 2}
                               If a GPU is not in this dict, it will use the default max_gpus value.
+            gpu_list: Optional list of GPU names to evaluate. If None, evaluates all GPUs in GPU_SPECS.
             max_ttft: Maximum time to first token constraint (ms)
             max_itl: Maximum inter-token latency constraint (ms)
             max_latency: Maximum end-to-end latency constraint (s)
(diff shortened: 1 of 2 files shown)


config_explorer/src/config_explorer/recommender/recommender.py
```diff
@@ -0,0 +1,199 @@
...
+                gpu=gpu_name,
+                num_gpus=self.max_gpus,
+                framework="vllm",
+                target="throughput",
+                constraints=constraints,
+            )
+
+            try:
+                updated_params, result = run_performance_estimation(params)
+                best_config = result.best_configs[0] if isinstance(result.best_configs, list) else result.best_configs
```
namasl: This line isn't being used for anything

config_explorer/src/config_explorer/recommender/recommender.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def get_gpu_with_highest_throughput(self) -> Optional[Tuple[str, float]]:
+        """
+        Get the GPU with the highest throughput from results.
+
+        Returns:
+            Tuple of (gpu_name, throughput_value) or None if no valid data
+        """
+        if not self.gpu_results:
+            return None
```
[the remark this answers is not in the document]
namasl: Likewise for other methods.

config_explorer/src/config_explorer/recommender/recommender.py
```diff
@@ -0,0 +1,199 @@
...
+from llm_optimizer.performance import PerformanceEstimationParams, PerformanceEstimationResult, run_performance_estimation
+
+class GPURecommender:
+    """Recommends optimal GPU for running LLM inference using BentoML's llm-optimizer roofline algorithm.
+
+    Given a list of models and available GPUs, recommends the best GPU
+    for each model based on synthetic performance estimates.
+    """
+
+    def __init__(
```
namasl: Does it make sense to have the set of GPU models as an init parameter for this class? During init the `recommend_gpu` class can run to get results right away.

We may also want `max_gpus` to be a per-GPU parameter. Maybe a user has to choose between 8 A100 and 4 H100.

jgchn: Screenshot 2026-01-05 at 10 26 43

namasl: Yes, that's what I'm suggesting. Have it as a parameter to `__init__()`, then store it as an instance variable (`self.gpu_list = gpu_list`).

PR #587, opened 2026-01-05
Reorg to allow multiple versions of BenchmarkReport schema

namasl:
This PR does a bit of reorganization in preparation for allowing the use of multiple `BenchmarkReport` schema versions.

There is a small placeholder for the version `0.2` `BenchmarkReport` which will be filled out in a future PR.


namasl:
Doing a different reorg...

PR #609, opened 2026-01-14
[WVA] Additional Documentation on Running Experiments

Vezio:
Enhances the existing documentation to provide a real example of running experiments with and without `wva`.


maugustosilva (approved):

maugustosilva (changes requested):
Take a look at line 104. The "broken link detector" is know to (falsely) detect long URLs as "broken"

maugustosilva (approved):

PR #635, opened 2026-01-30
add random concurrent workload template for inference-perf

huaxig:
introduces a new inference-perf workload profile for sanity testing with concurrent users.

The sanity_random_concurrent profile is designed to test the stability and performance of the llm-d stack under increasing load. It consists of 4 stages, starting with 1 concurrent user and scaling up to 8.


huaxig:
SG, rename it to `random_concurrent.yaml.in`

PR #664, opened 2026-02-12
Update benchmark report README

namasl:
Add details on benchmark report v0.2


benchmark_report/br_v0_1_example.yaml
```diff
@@ -0,0 +1,276 @@
...
+            max: 49.2
+        power:
+          units: Watts
+          mean: 410.02
+          stddev: 170.1
+          min: 201.3
+          p10: 243.4
+          p50: 314.3
+          p90: 475.1
+          max: 497.2
```
jgchn: Curious how this is measured?

jgchn (approved):

benchmark_report/br_v0_1_example.yaml
namasl: We don't have a way to do this in llm-d-benchmark right now, but here's an example of measuring energy consumption https://www.johnsnowlabs.com/tokens-per-joule-how-to-quantify-and-reduce-the-energy-footprint-of-clinical-llm-inference/#how-to-measure-benchmark-llm-energy-use

PR #711, opened 2026-02-20
Update config explorer tests

jgchn:
Fix config explorer failing tests. Root cause is that Qwen/Qwen-0.6B and llama-8B no longer returns safetensor metadata.


kalantar (approved):
/lgtm

PR #854, opened 2026-03-20
Fix metrics scrape

mengmeiye:
1. Remove redundant code which originally had two metric monitoring systems; Now only one
2. Fix the bug where metrics monitoring is only enabled when `-f` is there with `run.sh` instead of having it enabled as default
3. Plot metrics data as line or box graphs
4. Start adding metrics info into benchmark report v0.2 (but please wait for the next PR to achieve what #776 does)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -35,10 +35,6 @@ RUN apt-get update; \
 RUN curl -fsSL "https://dl.k8s.io/release/$(curl -fsSL https://dl.k8s.io/release/stable.txt)/bin/linux/$(dpkg --print-architecture)/kubectl" \
     -o /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
 
-# Install kubectl for in-pod cluster operations (e.g. vLLM metrics scraping)
-RUN curl -fsSL "https://dl.k8s.io/release/$(curl -fsSL https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" \
-    -o /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
-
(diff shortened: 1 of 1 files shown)


analysis/guidellm-analyze_results.sh
```diff
@@ -30,4 +30,27 @@ mkdir -p "$LLMDBENCH_RUN_EXPERIMENT_RESULTS_DIR/analysis"
...
+if [[ -f "$_metrics_dir/processed/metrics_summary.json" ]]; then
+  echo "Integrating metrics summary into benchmark report(s) v0.2..."
+  for _report in $(find "$LLMDBENCH_RUN_EXPERIMENT_RESULTS_DIR" -maxdepth 1 -name 'benchmark_report_v0.2,_*.yaml'); do
+    python3 -c "
+import yaml, sys
+from benchmark_report.metrics_processor import add_metrics_to_benchmark_report
+report_file, metrics_dir = sys.argv[1], sys.argv[2]
+with open(report_file) as f:
+    br_dict = yaml.safe_load(f)
+br_dict = add_metrics_to_benchmark_report(br_dict, metrics_dir)
```
namasl: Noting that this function adds metrics to a benchmark report YAML without validation. It imports the YAML as a Python `dict`, appends metrics, then we dump that `dict` as a new YAML here.

It would be safer to create an instance of `BenchmarkReportV02` from this `dict` (to do full validation), then use the [`export_yaml()`](https://github.com/llm-d/llm-d-benchmark/blob/2515dbeee3022b6416e073e6d4ab790363f7ccff/benchmark_report/base.py#L175) method to create the new YAML file.

This doesn't need to be fixed here since it is a problem with the existing code, but just raising it for awareness.

PR #859, opened 2026-03-23
[Run] Add --repeat flag to repeat experiments N times with aggregation

### LINKED ISSUE #701: Expand benchmark run scripts to repeat the experiment N times to account for randomness
### Component

Run

Expand benchmark run scripts to repeat the experiment N times, then combine results, to account for randomness. 

### Proposed solution

Expand benchmark run scripts (`run_only.sh` and maybe also `run.sh`) to repeat the experiment N times (with different seeds where applicable). Combine results and report mean/std dev. Default can be N=1.

### Alternatives

_No response_

_No response_

jia-gao:
## Summary

Adds support for running experiments multiple times and aggregating results with mean/std dev to account for benchmark variability. Closes #701.

- **`existing_stack/run_only.sh`**: New `-R/--repeat N` flag (default: 1, also settable via `LLMDBENCH_HARNESS_REPEAT` env var). Each repeated run gets a unique experiment ID (`{uid}_{workload}_run{i}`). After all runs complete, calls the aggregation script automatically.
- **`analysis/aggregate_runs.py`**: New script that reads Benchmark Report v0.2 YAML files from repeated runs and produces `aggregated_summary.json` and `aggregated_summary.txt` with mean, std, min, max for all performance metrics.

Backward compatible: `--repeat 1` (default) produces identical behavior to today — no suffix, no aggregation.

### Example usage

```bash
# Run each workload 3 times and aggregate
./existing_stack/run_only.sh -c my_config.yaml --repeat 3

# Or via env var
LLMDBENCH_HARNESS_REPEAT=5 ./existing_stack/run_only.sh -c my_config.yaml
```

### Example aggregated output

```
Aggregated Benchmark Results (3 runs)
Run IDs: run1, run2, run3
==========================================================================================
Metric                                                        Mean        Std        Min        Max
------------------------------------------------------------------------------------------
latency.time_to_first_token.mean                            0.0337     0.0015     0.0320     0.0350
latency.time_to_first_token.p99                             0.0500     0.0020     0.0480     0.0520
throughput.output_token_rate.mean                        1065.2333    14.8514  1050.5000  1080.2000
==========================================================================================
```

## Scope / follow-ups

- This PR only modifies `existing_stack/run_only.sh`. Adding `--repeat` to `setup/run.sh` (the full-stack orchestrator) is left as a follow-up since it has a more complex execution model.
- Seed control for harnesses can be added as a follow-up (passed as env var to the benchmark pod).

## Test plan

- [x] Verify `--repeat 1` (default) produces identical results/directory structure to current behavior
- [x] Verify `--repeat 3` creates 3 result directories per workload with `_run1`, `_run2`, `_run3` suffixes
- [x] Verify aggregation script produces correct `aggregated_summary.json` and `.txt`
- [x] Verify `--repeat 0` and `--repeat abc` are rejected with error message

### DIFF
--- a/analysis/aggregate_runs.py
+++ b/analysis/aggregate_runs.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python3
+
+# Copyright 2025 The llm-d Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Aggregate benchmark results across repeated runs.
+
+Reads Benchmark Report v0.2 YAML files from multiple runs of the same
+experiment and produces an aggregated summary with mean, std dev, min,
+and max for key performance metrics.
+
+Usage:
+    python3 aggregate_runs.py \
+        --results-prefix /requests \
+        --harness inference-perf \
+        --stack llm-d-7b-base \
+        --run-ids 1742680000_workload1_run1 1742680000_workload1_run2 \
+        --output /requests/1742680000_workload1_aggregated
+"""
+
+import argparse
+import glob
+import json
+import math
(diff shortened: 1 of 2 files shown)


maugustosilva (approved):

rachelt44:
@jia-gao Thank you for the implementation. I want to run each of the benchmark experiments from the same stack state. That is, I need kv cache eviction between experiments, otherwise previous runs affect new runs, and the experiments become dependent on each other. I don't see this in the current implementation, is that correct? should I open a new issue to request this addition?

PR #881, opened 2026-03-31
fix: Gracefully handle missing model PVC dynamically in pod spec generator

adinilfeld:
Introduce a `harness.mountModelCache` configuration flag (defaulting to true) to control the inclusion of the `cache-volume` volume and volumeMount in the harness data-access pod. Update the CLI to automatically disable this flag in "run-only" mode (e.g., when `--endpoint-url` or `--config` is specified). This resolves an issue where the pod would get trapped in a pending state when running against an existing stack, as `model-pvc` is only created during the `standup` phase. 

I verified the changes using dry-run executions of the `plan`, `standup`, and `run` commands. Testing confirmed that the `cache-volume` is correctly included by default for standard deployments and successfully excluded in run-only mode, all while passing Pydantic schema validation without warnings.

Additionally increase the timeout, as provisioning the `workload-pvc` can take longer than 2 minutes.

### DIFF
--- a/llmdbenchmark/standup/steps/step_05_harness_namespace.py
+++ b/llmdbenchmark/standup/steps/step_05_harness_namespace.py
@@ -64,7 +64,6 @@ class HarnessNamespaceStep(Step):
             context, "06_pod_access_to_harness_data"
         )
         if pod_yaml:
-
             result = cmd.kube("apply", "-f", str(pod_yaml))
             if not result.success:
                 errors.append(


adinilfeld:
> Is there a way we can handle this at render time instead? 

Good point - I've refactored the change to use a new `harness.mountModelCache` config flag, which defaults to `true` except in "run-only" mode (e.g., when `--endpoint-url` or `--config` is specified).

Updated the PR description with some more information.

PR #888, opened 2026-04-02
fix: Configuration file concatenation bugs and Crane resolution fallback

adinilfeld:
Corrects file appendage errors in `render_specification.py` using `with_suffix()`. Extends `version_resolver.py` capabilities by integrating `crane ls` alongside the Skopeo and Podman fallbacks for tag verification.


Vezio (approved):
Pretty good to me

PR #918, opened 2026-04-09
add parser to replace model.name automatically

mengmeiye:
Add a parser to update model name and path automatically in the scenario yaml files instead of having hard-coded values.


Vezio:
This seems pretty good! Is the only `scenario` that uses this in `spyre` or are there others that might be good to update too?

Vezio (commented):
I will approve but maybe can be applied to other scenarios using `customCommand` like wide-ep

Vezio (dismissed):

Vezio (dismissed):

Vezio (approved):

PR #926, opened 2026-04-10
feat: implement queue-based async telemetry

jjk-g:
Add a telemtry provider interface and http implementation.

Supports publish json data to a http endpoint with corresponding api-key

### DIFF
--- a/llmdbenchmark/cli.py
+++ b/llmdbenchmark/cli.py
@@ -27,6 +27,8 @@ from llmdbenchmark.utilities.os.filesystem import (
     resolve_specification_file,
 )
 from llmdbenchmark.interface.commands import Command
+from llmdbenchmark.telemetry import init_telemetry, get_telemetry
+import getpass
 from llmdbenchmark.interface import plan, standup, teardown, run
 from llmdbenchmark.interface import smoketest as smoketest_interface
 from llmdbenchmark.interface import experiment as experiment_interface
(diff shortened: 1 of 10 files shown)


llmdbenchmark/cli.py
```diff
@@ -1528,6 +1530,32 @@ def cli() -> None:
...
+            "time": int(time.time() * 1000),
+            "command": args.command,
+            "config": {
+                "specification_file": str(args.specification_file),
+                "workspace": str(config.workspace),
+                "dry_run": config.dry_run,
+                "verbose": config.verbose,
+            },
+            "environment": {
+                "LLMDBENCH_BASE_DIR": os.environ.get("LLMDBENCH_BASE_DIR"),
```
Vezio: I believe the base dir is eventually squashed into a dict. We accept `LLMDBENCH_BASE_DIR` as an environment variable override, but we take it in as a `cli` as well, otherwise this looks good to me here.

I wonder if the telemetry hooks would fit in with the `llmdbenchmark` top level CLI options like we have for workspace, etc... and we can similarly have them used as overrides via `env` if needed. Thoughts?

Vezio:
Looks like a pretty good interface. I added `Do Not Merge` tag for now just to give some time for people to review and not merge prematurely.

I have laid out one comment about entrypoint (env vs cli).

It's a nice implementation that hides a lot of the complexity from the caller, I like it, and conforms to existing patterns set in the code base.

Thanks for the prototype here @jjk-g

llmdbenchmark/cli.py
jjk-g: Updated params to have cli arg values.
Kept API Key as env only to hopefully prevent leaking it.
Updated base dir to access the args value instead of direct from env.

jjk-g:
Added bearer_token interface which was needed to work with our setup.

Adds google-auth dependency, necessary for oicd token retrieval for LLMDBENCH_TELEMETRY_AUTH_PROVIDER=google.

maugustosilva (approved):

Vezio (approved):

Issue #938, opened 2026-04-15
Add a linting process to format the codebase


Luka-D:
### Component

I don't know

### Desired use case or feature

Currently, various files in the codebase have different formatting styles. This makes contribution more difficult as auto-formatters have to be disabled or modified. 

### Proposed solution

Ideally, we can integrate a code linter and formatter as part of the pre-commit harness. Then users could run pre-commit before commiting code and have their code linted and/or formatted to match the rest of the repository. I think `pylint` was being discussed as the linter to move to. The Black formatter is a popular formatter that can be used in conjunction with pylint. Would we want the code to only be linted and rely on the users to change their formatting, or do we want to integrate an auto-formatter as well to automatically format users code?

### Alternatives

Alternatively: the linting and autofomatting process could also be moved to part of the PR checks process, similar to the existing markdown linter.

Luka-D:
Seems like another popular tool is [Ruff](https://docs.astral.sh/ruff/), which is both a linter and formatter.

Luka-D:
Process to integrate pylint into pre-commit: https://pylint.pycqa.org/en/stable/user_guide/installation/pre-commit-integration.html

Vezio:
`pylint` seems like where we want to go - I propose we start small and implement this as an `optional` thing to run - we will need to `clean up` a LOT of code as we run the scanners. Once we get to a place where the majority of the code base is conformant to `pylint` standards, then we can make it a mandatory thing.

@maugustosilva thoughts?

maugustosilva:
+1 on `pylint` and +1 on **optional** for starters. There will be A LOT of code cleanup to be done.

Luka-D:
## My investigation into this issue so far

I think we should implement both a code linter and a code formatter. 

A code linter would be helpful to determine bugs in the code and logical errors, but it does not automatically change the code. The user is still required to go back and manually change these errors. 

A code formatter is purely for stylistic consistency and does not impact the logic of the code. Currently, it seems to me that some files within the repo have different style standards than others in regards to spacing, line length etc. I think we should implement a code formatter as well so that the codebase has a general style, but we need to decide on which formatter to use.

### Choice of linter

[...]

PR #962, opened 2026-04-17
Feature: Added Ruff as a linter and formatter to pre-commit.

Luka-D:
Part of the issue #938 discussion, this is how we can integrate Ruff into the pre-commit workflow. Currently is is an optional step. You probably need to run `pre-commit install` first.

Then you can verify that the default `pre-commit` functionality is unchanged. You can make changes to some files, stage those changes and then run the following to lint and format:

```sh
pre-commit run ruff-check --hook-stage manual  # Lint check

pre-commit run ruff-format --hook-stage manual # Format code
```

If you want to lint and format the whole codebase, you can run:
```sh 
pre-commit run ruff-check --hook-stage manual  --all-files

pre-commit run ruff-format --hook-stage manual --all-files
```

I have code ready to add Ruff to the CI/CD pipeline as well whenever we feel ready for that. Let me know your thoughts! :)


Vezio:
Can you tie this in as dependency looks like users will need to install `ruff` - we have `install.sh` and as well as various `pre-commit` scripts that handle this, please take a look so we don't require any manual steps. Thanks!

Luka-D:
Hey Tyler, users don't need to manually install ruff themselves with `pip` or anything. Since it's added as a remote repo, `pre-commit` will automatically download ruff into a venv and run it there. All users need to do is run `pre-commit install` which is already part of the `./util/setup_precommit.sh` script. So I think we can ask all the devs to re-run that script and they should be good. I can modify the text in `setup_precommit.sh` to reflect the new addition of two pre-commit hooks.

Luka-D:
Upon further investigation, pre-commit does not install hook repositories when running `pre-commit install`, but rather upon the first time the hook is ran. Since both of the ruff hooks are manual, the first time a user manually runs them using `pre-commit run ruff-check --hook-stage manual`, pre-commit will install ruff for them and then use the cached version for subsequent runs.

If we want to install hook repo environments within the script, we can add the `pre-commit install-hooks` command to the `./util/setup_precommit.sh` file. This will offload the time taken upon first running that hook to instead be downloaded in the setup script. So basically, the user will wait a little longer during the setup script instead of waiting a little longer during their first (manual at this time) commit. This is optional but I'm curious to hear your thoughts.

Vezio:
>If we want to install hook repo environments within the script, we can add the pre-commit install-hooks command to the ./util/setup_precommit.sh

yes - that is the entire purpose of that script.

For example, already there is:

```
echo "==> Installing pre-commit framework and dev extras..."
pip3 install --upgrade pip
pip3 install -r .pre-commit_requirements.txt

# ---------------------------------------------------------------------
# 3. Register the pre-commit hook only. We deliberately don't register
#    a pre-push hook -- CI runs the full per-spec validation, so the
#    push-time gate would just duplicate work.
# ---------------------------------------------------------------------
echo ""
echo "==> Registering pre-commit hook..."
pre-commit install
```

Luka-D:
Sounds good! I've modified the script to now install the pre-commit hook environments when ran, it takes a few seconds on my machine from a clean pre-commit install.
image

Vezio (approved):

PR #963, opened 2026-04-17
[Optional] Uv install

mengmeiye:
This PR is for the case when there's no python virtual environment detected or the system Python is missing or older than `3.11`. User will be prompted whether to use [`uv`](https://docs.astral.sh/uv/) or the standard `python3 -m venv`. `uv` can automatically download the correct Python version if the system Python is missing or older than `3.11`.


Vezio:
@mengmeiye thanks for this - can you mention this enhancement in the `README` at the top level where we talk about installing the library. It might be good advertising that we support both `uv` and `venv` out of the box.

mengmeiye:
> @mengmeiye thanks for this - can you mention this enhancement in the `README` at the top level where we talk about installing the library. It might be good advertising that we support both `uv` and `venv` out of the box.

@Vezio yes, will do 🫡

mengmeiye:
> @mengmeiye thanks for this - can you mention this enhancement in the `README` at the top level where we talk about installing the library. It might be good advertising that we support both `uv` and `venv` out of the box.

@Vezio : Done ✅

Vezio (approved):
lgtm!

PR #964, opened 2026-04-17
fix the bugs for configMap under sidecar and preprocess script

mengmeiye:
- When `sidecar.enabled: true`, the chart dereferences `.Values.inferenceExtension.sidecar.configMap.name` to create a Kubernetes `configMap`. If `configMap` is not in the values, Helm hits a nil pointer error:
```bash
Error: template: inferencepool/templates/inferenceextension.yaml:1:3: executing
  "inferencepool/templates/inferenceextension.yaml" at <include "inference-extension.config" .>: error calling include:
  template: inferencepool/charts/inferenceExtension/templates/_config.yaml:63:18: executing "inference-extension.config"
  at <.Values.inferenceExtension.sidecar.configMap.name>: nil pointer evaluating interface {}.name
```

- Fix the bug for the undefined var name - `executable_actual_path` in `llmdbenchmark/standup/preprocess/set_llmdbench_environment.py`


Vezio (approved):
Lgtm

PR #986, opened 2026-04-19
fix bug when harness waitTimeout is defined in scenario

mengmeiye:
When `waitTimeout` is defined for `harness` in scenario, cli didn't really pick it up.
Example: https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/examples/spyre.yaml#L408
```yaml
harness:
      name: inference-perf
      experimentProfile: sanity_random.yaml
      waitTimeout: 36000
```
This PR resolved the issue.


Vezio (approved):
Thanks for fixing this!

PR #1018, opened 2026-04-23
Re-enabled all CI/CD workflows

maugustosilva:
Re-factored all workflows to nightly test on OpenShift (OCP), Coreweave (CKS) and Google (GKE).

In order to save time, only one harness is tested per cloud provider (the one specified on the cicd scenario)


Vezio (approved):
Thank you @maugustosilva for these changes it'll be amazing to have a full CICD enabled for real workloads nightly. THANK YOU

PR #1053, opened 2026-04-30
Adding deletion of requester plus timeout for deletion

aavarghese:


llmdbenchmark/teardown/steps/step_01_uninstall_helm.py
```diff
@@ -127,6 +141,23 @@ def _delete_fma_crs(
...
+        ]:
+            context.logger.log_info(
+                f"Waiting for FMA {label} to terminate in {namespace}..."
+            )
+            cmd.kube(
+                "wait", "pod",
+                "--for=delete",
+                f"--selector={selector}",
+                "--namespace", namespace,
+                "--timeout=120s",
```
Vezio: It might be good to parameterize the timeout in the function call so we don't have to hard code it here - I know MANY other places that have it hardcoded but were moving away from that. Otherwise LGTM

aavarghese: sure - good idea. Will add now

Vezio (approved):

PR #1103, opened 2026-05-05
feat: expose model_id_label as ${model.idLabel} for use in extraObjec…

### LINKED ISSUE #1102: Referencing InferencePool name in extraObjects
### Component

Setup/Standup

I would like to deploy an `Inference`Objective as part of my `llmdbenchmark standup`.

In theory, I can do this by specifying it in the `extraObjects` (see [https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/values/defaults.yaml#L1296) ](https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/values/defaults.yaml#L1296) field of the scenario. This gets added to the modelservice values file (see [https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/jinja/13_ms-values.yaml.j2#L1021-L1027) ](https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/jinja/13_ms-values.yaml.j2#L1021-L1027) and will be deployed by the modelservice helm chart.

However, an `InferenceObjective` references an `InferencePool` by name (`spec.poolRef.name`). The name of the `inferencePool` is defined (in the GAIE chart) as `Release.Name` (see: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/v1.4.0/config/charts/inferencepool/templates/inferencepool.yaml#L5C9-L5C28). In turn, this seems to be derived from `model_id_label`.  (See: https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/jinja/10_helmfile-main.yaml.j2#L54)

Is there an appropriate way to express, in the scenario, that the name is `model_id_label`?

### Proposed solution

@Vezio said (in slack):
Today there's no clean dotted-path that exposes model_id_label for use in `extraObjects`, the `${dotted.path}` substitution requires at least one dot in the reference (so it doesn't collide with shell vars like `${VLLM_PORT}` in `customCommand` strings).

The cleanest fix is probably to expose it as `${model.idLabel}` (which is about a one-line change in `_resolve_model_id_label` to alias it under model.). Then your `InferenceObjective` would look something like:

```
  spec:
    poolRef:
      name: ${model.idLabel}-gaie
```

### Alternatives

_No response_

_No response_

kalantar:
Fixes https://github.com/llm-d/llm-d-benchmark/issues/1102

Aliases the computed model_id_label into the model dict so the existing ${dotted.path} substitution mechanism can resolve it. This lets scenarios reference it, using `${model.idLabel}`.

### DIFF
--- a/llmdbenchmark/parser/render_plans.py
+++ b/llmdbenchmark/parser/render_plans.py
@@ -605,6 +605,8 @@ class RenderPlans:
         else:
             values["model_id_label"] = model.get("shortName", "")
 
+        model["idLabel"] = values["model_id_label"]
+
         return values
 
     # Defaults whose "bare" form collides across multi-stack scenarios -
--- a/tests/test_config_variable_substitution.py
+++ b/tests/test_config_variable_substitution.py
@@ -128,3 +128,30 @@ class TestSubstituteConfigVariables:
         }
         result = renderer._substitute_config_variables(values)
         assert result["field"] == "max len is 32768"
+
+
+class TestModelIdLabelAlias:
+    """Tests that _resolve_model_id_label exposes model_id_label as model.idLabel."""
+
+    def test_model_id_label_available_as_dotted_path(self, renderer):
+        """${model.idLabel} resolves to the computed model_id_label value."""
+        values = {
+            "model": {"name": "meta-llama/Llama-3.2-1B-Instruct"},
+            "namespace": {"name": "benchmark"},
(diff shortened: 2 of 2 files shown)


kalantar:
Manually tested this in a scenario. The inferenceObjectives were deployed and correctly reference the inferencepool.

maugustosilva (approved):

PR #1367, opened 2026-05-21
Add nixl benchmark script

praveingk:
**How to test:** 

On Prefill Pod: 
```
python benchmark_nixl.py --role server --device gpu --local-gpu-idx 0 --op-type read --backend ucx
```

On Decode Pod: 
```
python benchmark_nixl.py --role client --remote-ip <IPADDRESS> --device gpu --local-gpu-idx 0 --op-type read --backend ucx

```

This will report the results for various message sizes. 

@maugustosilva We will need to invoke this script from the harness on the pods. This script uses the existing installation on the pods, and doesn't require any other additional libraries to be installed.

### DIFF
--- a/llmdbenchmark/standup/preprocess/benchmark_nixl.py
+++ b/llmdbenchmark/standup/preprocess/benchmark_nixl.py
@@ -1,0 +1,767 @@
+from __future__ import annotations
+
+import argparse
+import sys
+import time
+from typing import List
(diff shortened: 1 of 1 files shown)


maugustosilva:
Thank you very much @praveingk !

maugustosilva (approved):

PR #1415, opened 2026-05-28
[Run] Add all guide-specific workloads on this repo

maugustosilva:
As part of the integration process between `llm-d` and `llm-d-benchmark` (where benchmarks are run nightly), all "guide-specific" workloads were copied here.


Vezio:
lgtm - maybe remove the suffix `1` and provide better name?

Vezio (dismissed):

PR #1436, opened 2026-05-31
Add tolerations and affinity support for inferenceExtension (EPP) pod

yardenmaymon-td:
## What

Propagate optional scheduling fields for the EndpointPicker (EPP) pod into the inferencepool chart values (`config/templates/jinja/12_gaie-values.yaml.j2`):

- **`inferenceExtension.tolerations`** — emitted only when a scenario sets a non-empty list, leaving the chart's `tolerations: []` default untouched otherwise.
- **`affinity`** — reuses the scenario-wide `affinity` convenience wrapper already consumed by `decode`/`prefill`/`standalone`: a flat `affinity.nodeSelector` map is expanded into a required `nodeAffinity` term, with `podAffinity`/`podAntiAffinity` passed through. Gated on `affinity.enabled`.

This lets the EPP pod be scheduled onto tainted nodes (e.g. GPU nodes) and steered toward specific nodes.

## Why these fields

Scoped to what the inferencepool/epplib chart (v1.4.0 and v1.5.0) actually consumes in its EPP pod spec — it honours `tolerations` and `affinity` but has **no `nodeSelector`** field, so propagating a top-level `inferenceExtension.nodeSelector` would be a silent no-op.

Note: the pod(anti)affinity passthrough uses `indent(_, true)` so the first rendered line is indented; the unparametrised `indent(6)` used by the modelservice template would leave the first line at column 0 (only hidden there because those keys default to `{}`).

## Testing

- New `tests/test_gaie_values_render.py` — renders the template with and without each field, including a pod-anti-affinity case that locks the first-line indentation. Values are based on `defaults.yaml` so the test stays resilient to unrelated template churn.
- Full suite: `pytest tests/` → 279 passed, 4 skipped.
- `cicd/kind` canary renders cleanly (29 files, 0 errors).

### DIFF
--- a/config/templates/jinja/12_gaie-values.yaml.j2
+++ b/config/templates/jinja/12_gaie-values.yaml.j2
@@ -51,6 +51,33 @@ inferenceExtension:
   resources:
 {{ inferenceExtension.resources | toyaml | indent(4, true) }}
 {% endif %}
+{% if inferenceExtension.tolerations is defined and inferenceExtension.tolerations %}
+  tolerations:
+{{ inferenceExtension.tolerations | toyaml | indent(4, true) }}
+{% endif %}
+{% if affinity is defined and affinity.enabled %}
+  affinity:
+{% if affinity.nodeSelector is defined and affinity.nodeSelector %}
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


maugustosilva:
Hello @yardenmaymon-td. Can you fix the conflicts? Thanks!

PR #1439, opened 2026-06-01
Support configurable listener port for the per-stack Gateway

yardenmaymon-td:
## What

Adds an optional `gateway.listenerPort` config that renders a custom HTTP listener into the per-stack Gateway resource (`11_infra.yaml.j2`).

- **gke / agentgateway / istio** (and any custom istio class): when `gateway.listenerPort` is set, a default HTTP listener on that port is rendered into the `listeners` block. When unset, the block is omitted and behaviour is unchanged.
- **data-science-gateway-class**: this class pins the listener to port 443 (HTTPS + TLS terminate), so `listenerPort` is unsupported there. A new `raise` jinja global makes the template fail loudly with a clear message instead of silently producing an invalid Gateway.

## Why

Lets scenarios expose the Gateway on a non-default port without editing templates, while keeping the fixed-port classes safe.

## Notes

- Default behaviour (no `listenerPort`) is unchanged for every gateway class.
- A falsy value (`0`) is treated as unset.
- Documented the option in `defaults.yaml`.

## Testing

New `tests/test_gateway_listener_port.py` (16 tests) renders the real template for every gateway class, with and without `listenerPort`, asserts the data-science guard raises, and covers the `raise` global. Full suite: 288 passed, 4 skipped.


maugustosilva:
Hello @yardenmaymon-td. Can you fix the conflicts? Thanks!

Vezio (approved):

PR #1440, opened 2026-06-01
guides/test, agentic: make runs deterministic

ruocco:


mengmeiye (approved):
LGTM! Let's wait until CI/CD passes so we get it merged.

PR #1485, opened 2026-06-10
Update p/d validation to handle kustomize configs

rlakhtakia:
This change adds support for Kustomize deployments to the P/D disaggregation validator:

1. Updates the pod inspection selectors to dynamically target pods using the llm-d.ai/guide label, validating both the prefill and decode groups
2. Bypasses static scenario configuration checks to allow validation to succeed across various guide variants.

### DIFF
--- a/llmdbenchmark/smoketests/validators/pd_disaggregation.py
+++ b/llmdbenchmark/smoketests/validators/pd_disaggregation.py
@@ -37,7 +37,13 @@ class PdDisaggregationValidator(BaseSmoketest):
             or ""
         )
 
+        is_kustomize = "kustomize" in context.deployed_methods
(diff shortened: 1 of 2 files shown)


rlakhtakia:
@achandrasekar @liu-cong ptal.

llmdbenchmark/smoketests/validators/pd_disaggregation.py
```diff
@@ -37,7 +37,13 @@ def run_config_validation(
             or ""
         )
 
+        is_kustomize = "kustomize" in context.deployed_methods
```
achandrasekar, line 40: Nit: We might want to deduplicate this logic with base.py so we have this one place.

achandrasekar (approved):

PR #1490, opened 2026-06-11
Normalize BR v0.2 conversion when otel_traces sets concurrent_sessions=0

ruocco:
The v0.2 schema constrains LoadStandardized.concurrency to >=1. otel_traces.yaml under inference-perf legitimately accepts concurrent_sessions: 0 to mean "unbounded" -- i.e. allow as many concurrent sessions as there are total sessions.

When concurrent_session is set to zero force it to num_sessions.

### DIFF
--- a/llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
+++ b/llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
@@ -47,8 +47,6 @@ def _normalize_concurrency(value: Any, zero_fallback: Any = None) -> Any:
         as_float = float(value)
     except (TypeError, ValueError):
         return value
-    if as_float == float("inf"):
-        return value
     if as_float == 0:
         return zero_fallback
     return value
--- a/llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
+++ b/llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
@@ -38,6 +38,8 @@ def _normalize_concurrency(value: Any, zero_fallback: Any = None) -> Any:
     with ``concurrent_sessions: 0`` meaning "no limit") would fail
     Pydantic validation. Substitute ``zero_fallback`` for 0 -- callers
     that have a meaningful cap (e.g. ``num_sessions``) pass it in.
+    For callers that don't have a suitable cap, a default `zero_fallback`
+    of `None` will be used.
     Negative and non-numeric values are returned as-is so validation
     still rejects them.
     """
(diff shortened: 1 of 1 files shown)


llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
```diff
@@ -30,6 +30,30 @@
...
+    still rejects them.
+    """
+    if value is None:
+        return None
+    try:
+        as_float = float(value)
+    except (TypeError, ValueError):
+        return value
+    if as_float == float("inf"):
+        return value
```
namasl:
```suggestion
```
I don't think this `if` statement is doing anything useful.

llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
```diff
@@ -30,6 +30,30 @@
 from .schema_v0_2_components import HostType
 
 
+def _normalize_concurrency(value: Any, zero_fallback: Any = None) -> Any:
```
namasl, line 33: The default value of `zero_fallback` is `None`, so any `0` value will now be silently dropped rather than fail Pydantic validation. Is this the intended behavior?

ruocco: Yes. Zero _is_ a valid value, and that means "take the max number, i.e.  num_session"

llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
```diff
@@ -30,6 +30,28 @@
...
+    """Map upstream "0 = unbounded" sentinels for the v0.2 schema.
+
+    LoadStandardized.concurrency is constrained to >=1, so a literal 0
+    coming back from a harness (e.g. inference-perf trace_session_replay
+    with ``concurrent_sessions: 0`` meaning "no limit") would fail
+    Pydantic validation. Substitute ``zero_fallback`` for 0 -- callers
+    that have a meaningful cap (e.g. ``num_sessions``) pass it in.
+    Negative and non-numeric values are returned as-is so validation
```
namasl, line 43:
```suggestion
    that have a meaningful cap (e.g. ``num_sessions``) pass it in.
    For callers that don't have a suitable cap, a default `zero_fallback`
    of `None` will be used.
    Negative and non-numeric values are returned as-is so validation
```

llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
namasl: Thanks, can we add something like the comment I just suggested to the docstring to make this more explicit?

ruocco: I think your suggestion is already good enough, or should we be even more verbose?

namasl (approved):

llmdbenchmark/analysis/benchmark_report/native_to_br0_2.py
namasl: Good enough, I'll merge once the GitHub workflows pass. Thanks!

PR #1493, opened 2026-06-11
feat: pass routing.proxy.resources through to modelservice values

jgchn:
The llm-d-modelservice helm chart now supports resource requests/limits on the routing-proxy initContainer. Wire the field through the Jinja template so scenario authors can set CPU/memory for the sidecar.


mengmeiye:
Hi @jgchn can you sign your commit please? Thanks.

mengmeiye (approved):

Issue #1465, opened 2026-06-05
[Feature]: Tighter FMA timing via DPC log parsing


rubambiza:
### Component

Harnesses

### Desired use case or feature

The current FMA per-path timing metrics (T_hot, T_warm/T_instance_create, T_cold_launcher) use Kube timestamps as upper bounds, which include kubelet readiness probe delay (up to 10s). This inflates measurements, particularly for hot starts where the actual wake latency is ~1-3s but we measure 8-10s.

PR llm-d-incubation/llm-d-fast-model-actuation#522 added `httpCallStartTime` and `k8sCallStartTime` fields to DPC logs for all external calls at microsecond precision. This enables tighter per-request timing that excludes the kubelet probe tail.

### Proposed solution

Parse the DPC controller logs (already collected by `write_controller_log` at end of benchmark) to extract tighter timing:

1. **End instant for all paths:** Find "Successfully relayed the readiness" log line for each requester, extract `httpCallStartTime` (RFC3339Nano). This is when the DPC confirmed the requester accepted the `/v1/become-ready` POST.

2. **Start instant for T_instance_create (warm):** Find "Created vLLM instance" log with `httpCallStartTime` for the matching instance ID. The `httpCallStartTime` captures when the DPC issued `CreateNamedInstance`.

3. **Start instant for T_hot:** Find "Woke inference server" log with `httpCallStartTime`. This is when the DPC issued `/wake_up`.

4. **Start instant for T_cold_launcher:** Find "Created launcher-based server-providing pod" log with `k8sCallStartTime`. This is when the DPC issued the launcher pod creation API call.

5. **Correlation:** Match log lines to iterations by requester pod name (included in klog structured fields as `requesterName`).

**DPC log format (klog):**
```
I0603 15:29:43.683293  1 inference-server.go:1220] "Bound server-providing Pod" worker=0 requesterName="fma-requester-..." name="launcher-..." node="..." httpCallStartTime="2026-06-03T15:29:43.670001787Z"
```

### Alternatives

- Use `dual_label_timestamp` from the harness watch (captures DPC bind time, not actuation completion -- too early for warm/cold paths due to bind-first architecture in alpha.13)
- Prometheus histograms from DPC (only provides aggregate distributions, not per-request correlation)

### Additional context or screenshots

- Depends on: llm-d-incubation/llm-d-fast-model-actuation#522 (merged -- adds `httpCallStartTime`/`k8sCallStartTime` to DPC logs)
[...]

PR #1497, opened 2026-06-12
fix: Handle missing per-path timing fields in FMA analysis

### LINKED ISSUE #1496: [Bug]: FMA analysis crashes with KeyError on missing per-path timing fields
### Component

Analysis/Plotting

### Describe the bug

`nop-analyze_results.py` crashes with `KeyError: 'value'` when processing FMA benchmark results that contain a mix of actuation conditions (hot/warm/cold). The bug triggers on any run where at least one iteration lacks a particular per-path timing field (e.g., a warm-start iteration has no `t_wake`).

```
File "/usr/local/bin/nop-analyze_results.py", line 329, in write_fma_metrics
    t_hot_val = float(t_hot["value"]) if isinstance(t_hot, dict) else None
                      ~~~~~^^^^^^^^^
KeyError: 'value'
```

Root cause: Lines 328-333 use `.get("t_wake", {})` with an empty dict default. When the key is absent, the empty dict passes the `isinstance(t_hot, dict)` check, then `{}["value"]` raises KeyError.

### Steps to reproduce

1. Run an FMA benchmark with mixed actuation paths (e.g., iterations that include both hot and warm starts)
2. Wait for the analysis step to process `result.yaml`
3. The analysis script crashes because warm-start iterations lack `t_wake`, and hot-start iterations lack `t_instance_create`

Observed in: https://github.com/llm-d/llm-d-benchmark/actions/runs/27386829874

Fix: Change the `.get()` default from `{}` to `None` for `t_wake`, `t_instance_create`, and `t_cold_launcher` lookups (lines 328-333). Additionally guard the dict access with a `"value" in` check so that an unexpected dict structure does not crash the script.

Introduced by #1429. Blocks #1465 (DPC log parsing writes to the same fields).

rubambiza:
## Summary

Fixes #1496

When an FMA per-path timing field (e.g. `t_wake`, `t_instance_create`, `t_cold_launcher`) is absent from the launcher info, `.get("t_wake", {})` returns an empty dict. Since `isinstance({}, dict)` is `True`, the code proceeds to access `{}["value"]`, raising a `KeyError`.

The fix changes the `.get()` default from `{}` to `None` and adds a `"value" in` guard on the dict access, so missing fields gracefully resolve to `None` instead of crashing.

This unblocks #1465.

## Test plan

- Pre-commit hooks pass (pytest, ruff-format, ruff lint, detect-secrets)
- Run analysis on a results file that has iterations missing one or more of the timing fields and confirm no KeyError is raised

Assisted by Claude Code

### DIFF
--- a/llmdbenchmark/analysis/scripts/nop-analyze_results.py
+++ b/llmdbenchmark/analysis/scripts/nop-analyze_results.py
@@ -325,12 +325,24 @@ def write_fma_metrics(  # pylint: disable=too-many-locals,too-many-statements
             elif actuation_condition == "T_cold_launcher":
                 cold_starts += 1
 
-            t_hot = launcher_info.get("t_wake", {})
-            t_hot_val = float(t_hot["value"]) if isinstance(t_hot, dict) else None
-            t_warm = launcher_info.get("t_instance_create", {})
-            t_warm_val = float(t_warm["value"]) if isinstance(t_warm, dict) else None
-            t_cold = launcher_info.get("t_cold_launcher", {})
-            t_cold_val = float(t_cold["value"]) if isinstance(t_cold, dict) else None
+            t_hot = launcher_info.get("t_wake")
+            t_hot_val = (
+                float(t_hot["value"])
+                if isinstance(t_hot, dict) and "value" in t_hot
+                else None
+            )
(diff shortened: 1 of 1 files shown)


aavarghese (approved):
LGTM! Thanks @rubambiza.

PR #1535, opened 2026-06-22
skip gateway API installation in Openshift

aleskandro:
In Openshift, the Gateway API CRDs are managed by the Ingress Operator, and a manual update as done in this repo is denied. The first commit of this PR skips the installation of the Gateway API CRD when the platform is Openshift.

~~Moreover, new versions of the Gateway API inference extension define the inference pool as v1 in the group inference.networking.k8s.io. The second commit introduces the WVA Helm Chart's value poolGroup in the related Jinja template, with a default to the group `inference.networking.k8s.io`.~~

These changes have been tested with the WVA scenario in Openshift 4.22.2.

### DIFF
--- a/llmdbenchmark/standup/steps/step_02_admin_prerequisites.py
+++ b/llmdbenchmark/standup/steps/step_02_admin_prerequisites.py
@@ -117,6 +117,7 @@ class AdminPrerequisitesStep(Step):
 
         self._install_gateway_api_crds(
             cmd,
+            context,
             plan_config,
             errors,
             existing_crds,
@@ -235,6 +236,7 @@ class AdminPrerequisitesStep(Step):
     def _install_gateway_api_crds(
         self,
         cmd: CommandExecutor,
+        context: ExecutionContext,
         plan_config: dict,
         errors: list,
         existing_crds: list[str],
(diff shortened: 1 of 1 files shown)


aleskandro:
Hi @asm582, can you give a look at this PR?

Vezio:
This is a good contribution however we will be getting rid of the `WVA` installation via `helm` and moving to the `WVA` installation via `kustomize` - @aavarghese is working on this - could you two collaborate on integrating this fix?

Thanks!

aleskandro:
Hi @Vezio , thanks. I removed the commit about the Helm value and will coordinate with @aavarghese about it in a separate PR if you're ok with it.

maugustosilva:
Hey @aleskandro thanks for your contribution. My main problem with it, as is, is that we can indeed have newer`istio` versions even in `OpenShift` (by indicating that the `crds` will not be managed by `servicemesh`). Taking this into account, I believe `if context.is_openshift` is far too encompassing. 

I propose we solve this by reintroducing an "openshift" or "servicemesh" `gatewayclass`, which will then simply skip the `crd` installation in case of `istio`.

aleskandro:
Hi @maugustosilva, thanks for the review. In my test, it was the ingress cluster operator that refused to install the Gateway API CRDs. Service Mesh or Istio was not installed at that time. Maybe I'm missing something?

maugustosilva:
Hey @aleskandro, you might be right. I was testing in two different OpenShift clusters, a `4.18` and `4.20` and might have mixed up which components were really blocking the `crd` installation. Be as it may, simply saying "`if context.is_openshift`, then skip any `crd` installation" is just too generic for the needs of users/experiments which are testing/exploring multiple different gateway providers. Hence, the proposal of a new `gatewayclass` (e.g., "openshift")

PR #1537, opened 2026-06-22
fix: gracefully handle DRA-managed accelerators in auto-detection

xiaojun-zhang:
## Problem

The nightly **Optimized Baseline E2E (Intel XPU)** job started failing after #1511 flipped the default `acceleratorType.labelValue` to `auto`. 

Root cause: Intel XPU exposes GPUs via **Dynamic Resource Allocation** (a `ResourceClaimTemplate` + `DeviceClass gpu.intel.com`), not a `node.status.capacity` extended resource or a GPU SKU label. `ClusterResourceResolver` only scans node capacity and labels, so on a DRA-only cluster `auto` cannot be resolved and the section falls into the fatal `else`. Previously the hardcoded literal default meant detection never ran, so the same cluster passed.

### DIFF
--- a/llmdbenchmark/parser/cluster_resource_resolver.py
+++ b/llmdbenchmark/parser/cluster_resource_resolver.py
@@ -63,6 +63,10 @@ class NodeResources:
     # e.g. {"nvidia.com/gpu.product": ["NVIDIA-H100-80GB-HBM3"]}
     gpu_labels: dict[str, list[str]] = field(default_factory=dict)
 
+    # DRA driver names from ResourceSlices, e.g. ["gpu.intel.com"]. These
+    # accelerators expose no node capacity resource or SKU label.
+    dra_drivers: list[str] = field(default_factory=list)
+
 
 class ClusterResourceResolver:
     """Resolve ``"auto"`` cluster resource values by querying node capacities and labels.
@@ -284,7 +288,13 @@ class ClusterResourceResolver:
             resources.accelerator_resources = sorted(accel_set)
             resources.network_resources = sorted(net_set)
             resources.gpu_labels = {k: sorted(v) for k, v in gpu_labels.items()}
+            resources.dra_drivers = self._scan_dra_drivers()
 
+            if resources.dra_drivers:
+                self.logger.log_info(
+                    f"Discovered DRA accelerator drivers: "
+                    f"{', '.join(resources.dra_drivers)}"
+                )
             if resources.accelerator_resources:
                 self.logger.log_info(
                     f"Discovered accelerator resources: "
@@ -310,6 +320,28 @@ class ClusterResourceResolver:
 
         self._node_resources = resources
         return resources
+
+    def _scan_dra_drivers(self) -> list[str]:
+        """Driver names from ResourceSlices (resource.k8s.io). Empty if none/unsupported."""
+        from kubernetes import client as k8s_client
+
+        drivers: set[str] = set()
+        for version in ("v1", "v1beta2", "v1beta1"):
+            try:
+                api = k8s_client.CustomObjectsApi(self._api_client)
+                slices = api.list_cluster_custom_object(
+                    group="resource.k8s.io",
+                    version=version,
+                    plural="resourceslices",
+                )
+            except Exception:
+                continue
+            for item in slices.get("items", []):
+                driver = (item.get("spec") or {}).get("driver")
+                if driver:
+                    drivers.add(driver)
+            break  # first served version wins
+        return sorted(drivers)
 
     @staticmethod
     def _vendor_prefixes(accelerator_resources: set[str] | list[str]) -> set[str]:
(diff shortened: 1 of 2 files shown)


llmdbenchmark/parser/cluster_resource_resolver.py
```diff
@@ -63,6 +63,10 @@ class NodeResources:
     # e.g. {"nvidia.com/gpu.product": ["NVIDIA-H100-80GB-HBM3"]}
     gpu_labels: dict[str, list[str]] = field(default_factory=dict)
 
+    # DRA driver names from ResourceSlices, e.g. ["gpu.intel.com"]. These
+    # accelerators expose no node capacity resource or SKU label.
+    dra_drivers: list[str] = field(default_factory=list)
```
Vezio, line 70: Pretty good here to separate

llmdbenchmark/parser/cluster_resource_resolver.py
```diff
@@ -311,6 +321,28 @@ def _scan_nodes(
         self._node_resources = resources
         return resources
 
+    def _scan_dra_drivers(self) -> list[str]:
+        """Driver names from ResourceSlices (resource.k8s.io). Empty if none/unsupported."""
+        from kubernetes import client as k8s_client
```
Vezio: Not sure if we've been doing this in other PRs, but a I've been advocating that `all` imports be declared at the top most of all python files, just to keep it standard across.

llmdbenchmark/parser/cluster_resource_resolver.py
```diff
@@ -311,6 +321,28 @@ def _scan_nodes(
...
+        drivers: set[str] = set()
+        for version in ("v1", "v1beta2", "v1beta1"):
+            try:
+                api = k8s_client.CustomObjectsApi(self._api_client)
+                slices = api.list_cluster_custom_object(
+                    group="resource.k8s.io",
+                    version=version,
+                    plural="resourceslices",
+                )
+            except Exception:
```
Vezio, line 335: What is the sideaffect of nothing being found?

llmdbenchmark/parser/cluster_resource_resolver.py
```diff
@@ -311,6 +321,28 @@ def _scan_nodes(
...
+                continue
+            for item in slices.get("items", []):
+                driver = (item.get("spec") or {}).get("driver")
+                if driver:
+                    drivers.add(driver)
+            break  # first served version wins
```
Vezio: If v1 is served by the API server but returns 0 ResourceSlices (e.g., on a cluster that has the CRD registered but no actual DRA resources installed yet), we break out without trying v1beta2/v1beta1. In practice if v1 is served, v1beta versions either also exist (duplicate data) or this cluster genuinely has no DRA resources so this is fine. 

Worth a tiny comment maybe, but overall ok to me :)

Vezio (dismissed):
Nice fix. the new branch slots cleanly into the existing graceful-degradation chain and should unblock the Intel XPU nightly. 

I do have a non-blocking suggestion for a follow-up (or this PR if you want to bundle it):

The new `dra_drivers` list could also feed `_vendor_prefixes` so the SKU-label heuristic still has a chance to match on `DRA-only` clusters. 

Today, on a hypothetical Intel cluster that exposes GPUs via DRA and labels nodes with gpu.intel.com/family=dg2:

```
accelerator_resources = []                   # DRA, no capacity entry
accel_vendors         = set()                # derived from empty list
dra_drivers           = ["gpu.intel.com"]    # this PR
```

`_looks_like_gpu_sku_label("gpu.intel.com/family", set())` will reject the label at the vendor gate (no prefix matches an empty set), so `gpu_labels` stays empty, and the new branch drops the `acceleratorType` constraint entirely. 

That's correct, but strictly less precise than what's available, on a heterogeneous Intel cluster we lose SKU isolation when the label was right there.

A one-line tweak in `_scan_nodes` should probably cover it:

accel_vendors = self._vendor_prefixes(accel_set) | {
    f"{d}/" for d in resources.dra_drivers
}

After that, the same cluster resolves to `labelKey=gpu.intel.com/family, labelValue=dg2`, and the new DRA fallback becomes a true last resort (DRA detected, but no SKU label of any shape) rather than the first thing that fires.

---

I'm going to approve this PR as is but just wanted to highlight this as a future opportunity to refine this in a case where we have this situation

llmdbenchmark/parser/cluster_resource_resolver.py
xiaojun-zhang: yes, it was doing this in _scan_nodes() methods. Totally agree it's better to keep it standard across. Will move both imports to the top.

Vezio: Thanks @xiaojun-zhang I appreciate that, we'll just need to keep enforcing stronger in other PRs, I've fallen victim as well 😂 - usually having the imports in each method directly makes writing local tests for that function easier since it's all self contained, but in the production case for our packaging, it makes better sense to have it at the top.

llmdbenchmark/parser/cluster_resource_resolver.py
xiaojun-zhang: in that case, it fallback to `acceleratorType.labelValue` as `auto`.

Vezio (approved):

PR #1556, opened 2026-06-25
fix: use priority-based GPU label selection for deterministic affinity

aleskandro:
## Summary

  - Implement priority-based GPU label selection to prefer more specific labels (e.g., product) over generic ones (e.g., family, class)
  - Add deterministic fallback when multiple GPU labels are available on nodes
  - Fix non-deterministic GPU label selection that previously used next(iter())

## Changes

  - Added GPU_LABEL_PRIORITY class constant defining preference order: cloud provider labels ? product ? product-name (AMD) ? family ? class
  - Implemented _select_best_gpu_label_key() method to select the most appropriate label based on priority ordering
  - Updated _resolve_affinity_node_selector() and _resolve_accelerator_type() to use the new priority-based selection

## Motivation

  Previously, when multiple GPU labels were present on a node (e.g., nvidia.com/gpu.product and nvidia.com/gpu.family), the code used next(iter()) to select one arbitrarily. This was non-deterministic and could result in less
  specific labels (like family) being chosen over more specific ones (like product), leading to incorrect or overly broad node affinity rules. This affected the WVA tests, still relying on the VariantAutoscaling object's annotation `acceleratorName`.

### DIFF
--- a/llmdbenchmark/parser/cluster_resource_resolver.py
+++ b/llmdbenchmark/parser/cluster_resource_resolver.py
@@ -102,6 +102,18 @@ class ClusterResourceResolver:
         }
     )
 
+    # Priority order for GPU label attribute selection when multiple labels
+    # are available. Labels are matched by their final attribute name (after
+    # the last `/` or `.`). Cloud provider labels (no attribute suffix) are
+    # treated as highest priority. Earlier items are preferred.
+    GPU_LABEL_PRIORITY = [
+        None,  # Cloud provider labels (cloud.google.com/gke-accelerator)
+        "product",  # Most specific SKU identifier
+        "product-name",  # AMD's variant of product
+        "family",  # GPU family (less specific than product)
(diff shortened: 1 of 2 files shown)


aleskandro:
Hi @Vezio, I've seen some recent commits of yours related to the discovery code. Could you review this PR if you get the chance?

Vezio:
@aleskandro I'm okay with this - main motivation for this is for those to just standup infra without having to worry about GPU resources - but tbh - for cases where those resources matter, a user should clearly specify those rather than relying on `auto` which is inherently non-deterministic on heterogeneous compute clusters.

I'm good with this PR to make those case more deterministic, but I still think with `auto` there will be non-deterministic configurations, unless a user performing said experiment or standup explcitly calls out the resource they're benchmarking against.

Vezio (approved):
I'll approve but will only merge after CICD passes

@aleskandro thanks for this!

PR #1564, opened 2026-06-26
Improving FMA + WVA benchmarking 

dumb0002:
This PR provides some fixes that improve the results computed by FMA + WVA benchmarking. 

  **Proposed Changes**

  - :gift: Default the WVA+FMA nightly workflow to `shared_prefix_synthetic_heavy.yaml` (was shared_prefix_synthetic.yaml). The heavy profile is calibrated to exceed 1-replica capacity so WVA's saturation engine actually triggers scaling — the light profile didn't exercise the autoscaling path under test.
  - :broom: Bump FMA pin to `v0.6.0` (was `v0.6.0-alpha.13`) across CRDs, chart version, and the three image tags (image.tag, requester.image.tag, launcher.image.tag) in config/templates/values/defaults.yaml. v0.6.0 is the stable release that promoted v0.6.0-alpha.14.
  - :bug: Fix the FMA requester Deployment's nodeAffinity in `config/templates/jinja/24_fma-deployment.yaml.j2` to use operator: In with a resolved value (was operator: Exists). WVA's accelerator resolver extracts the value from `matchExpressions[*].values[0]` for In operators, but returns empty for Exists. The empty value caused WVA's saturation engine to emit AcceleratorNotResolved + OptimizationFailed: No saturation metrics available for model for the entire run, leaving FMA scaling purely reactive (and too late). The template now reads the resolved value from `decode.acceleratorType.labelValue` (auto-resolved by`cluster_resource_resolver.py`), with prefill/standalone fallbacks and a `NVIDIA-H100-80GB-HBM3` final fallback.
  - :bug: Align vLLM configuration across baseline decode and FMA launcher so the benchmark measures scaling/serving differences, not vLLM-flag asymmetry:
    - Add `--no-enable-prefix-caching` to baseline decode.vllm.additionalFlags (FMA's launcher hard-codes this; baseline was getting ~16× free prefill on the shared-prefix workload because of 32-way prefix reuse per group).
    - Set `enforceEager: false` on both scenarios (was true in vllmCommon.flags). Baseline was running without CUDA graphs while FMA's launcher (whose options don't read this flag) had them enabled — measurable ~10-30% throughput bias against baseline. Both now use CUDA graphs.

  **Files changed**
  
-   `.github/workflows/ci-nightly-benchmark-ocp-fma-wva.yaml`         ← heavy workload as default
-   `config/scenarios/cicd/ocp-nofma-wva.yaml`                          ← --no-enable-prefix-caching, enforceEager: false
-   `config/scenarios/guides/workload-autoscaling.yaml`                 ← enforceEager: false
-   `config/templates/jinja/24_fma-deployment.yaml.j2`                  ← nodeAffinity operator: Exists → In
-   `config/templates/values/defaults.yaml`                             ← FMA v0.6.0-alpha.13 → v0.6.0
  
**Improved Results:**
Screenshot 2026-06-26 at 3 10 22 PM

### DIFF
--- a/.github/workflows/ci-nightly-benchmark-ocp-fma-wva.yaml
+++ b/.github/workflows/ci-nightly-benchmark-ocp-fma-wva.yaml
@@ -1,9 +1,13 @@
 name: CI - Nightly Run Benchmark on OCP ("WVA+FMA" autoscaling path)
 
 # Mirrors ci-nightly-benchmark-ocp-fma.yaml but targets the WVA(modelservice)+FMA scenario
-# under inference-perf load. Default harness
-# is inference-perf + shared_prefix_synthetic.yaml (same as running
-# guides/workload-autoscaling.yaml manually).
+# under inference-perf load. Default harness is inference-perf +
+# shared_prefix_synthetic_heavy.yaml — calibrated to drive WVA's saturation
+# gauges past the optimizer's threshold so HPA actually scales out during
+# the run. The light shared_prefix_synthetic.yaml profile peaks at 20 RPS,
+# which doesn't saturate even one Qwen3-32B replica, so neither path
+# exercises autoscaling and FMA pays its steady-state overhead with no
+# scaling benefit to amortize it. Override via inputs.workload if needed.
 #
 # This workflow runs TWO benchmark passes for direct comparison:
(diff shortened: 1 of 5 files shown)


aavarghese (approved):
LGTM! Thank you!!

PR #1498, opened 2026-06-12
feat: FMA + WVA benchmarking via workload-autoscaling guide

aavarghese:
Adds a manually dispatchable CI workflow that runs two benchmark passes for direct
comparison:

1. **Baseline** — `cicd/ocp-nofma-wva`: WVA scaling a `modelservice` decode Deployment.
2. **WVA + FMA** — `guides/workload-autoscaling`: same gateway/EPP/InferencePool path,
   but FMA layered on top — launcher pods host vLLM, the requester Deployment is the
   WVA scale target.

## Key changes

1. **FMA as a feature flag parallel to `wva.enabled`**.
   `modelservice.enabled: true, fma.enabled: true` renders and deploys both stacks.

2. **WVA install: kustomize via upstream
   `llm-d-workload-variant-autoscaler/config/overlays/namespace-scoped/openshift`**
   (remote git URL, no clone).

3. **WVA migrated to the modern annotation-based HPA discovery**
   No more `VariantAutoscaling` CR; HPA carries
   `llm-d.ai/managed=true`, `llm-d.ai/model-id`, `llm-d.ai/variant-cost`. External
   metric selector: `variant_name=<HPA name>`, `exported_namespace=<ns>` —
   `controller_instance` selector dropped.

4. **Cluster-state capture** (`step_09a_capture_cluster_state`) — snapshots HPA
   describe, WVA controller log, pod logs, events, prom-adapter, external-metrics
   API, Thanos diagnostic queries to `results/<exp_id>/`.

5. **Cold-start mean (s)** in the comparison table is computed from the WVA
   controller log alone — pairs each `"shouldScaleUp":true` decision with the next
   reconcile reporting `"totalReplicas":N+1`.`T_actuation mean (s)` averages 
   create→Ready across pods that reached Ready=True.

## Test plan

- [x] Local validation on pokprod (fma-ansu): standup → run → teardown for both
      `cicd/ocp-nofma-wva` and `guides/workload-autoscaling`
- [x] Both passes scale via WVA+HPA (`shouldScaleUp:true` events, HPA SuccessfulRescale)
- [x] FMA-only nightly (`ci-nightly-benchmark-ocp-fma.yaml`) keeps passing unchanged

```bash
# Baseline
llmdbenchmark --spec cicd/ocp-nofma-wva standup -p <ns>
llmdbenchmark --spec cicd/ocp-nofma-wva run     -p <ns>
llmdbenchmark --spec cicd/ocp-nofma-wva teardown -p <ns>

[...]

PR #1566, opened 2026-06-28
fix: query Deployment (not ReplicaSet) in find_fma_endpoint

aavarghese:
PR #1498 changed the FMA requester resource in `24_fma-deployment.yaml.j2` from `kind: ReplicaSet` to `kind: Deployment`, but `find_fma_endpoint()` in `endpoint.py` was left querying `replicaset`.  On the FMA-only nightly (`--methods fma`) the query returns nothing, `service_ip` stays None, and step_03 fails with `"Could not detect endpoint"`.

Fix: try Deployments first, then fall back to ReplicaSets for any cluster that was stood up before the change was deployed.


dumb0002:
LGTM! Thanks

dumb0002 (approved):

PR #1504, opened 2026-06-15
feat: Tighter FMA timing via DPC log parsing

rubambiza:
## Summary

- Add `dpc_log_parser.py` module that parses DPC controller logs to extract
  `httpCallStartTime`/`k8sCallStartTime` fields at microsecond precision
- Integrate parser into `benchmark_fma()` to override Kube-timestamp upper
  bounds with tighter DPC-derived values when available
- Graceful fallback: if DPC log data is unavailable (pre-PR llm-d-incubation/llm-d-fast-model-actuation#522 DPC, low
  verbosity, requester crash before relay), Kube upper bounds stand unchanged
- Streams log files line-by-line to handle large DPC logs without memory pressure
- Propagate `dpc_timing_available` through `native_to_br0_1` conversion so the
  analysis report attributes each row to its true source
- Adds "Source" column (DPC/Kube) to analysis output

Closes #1465

## Timing improvement (validated on cluster)

Measured on a 10-iteration `nop` run (scale 0->1->0 per iteration). The DPC log
parser now yields sub-second hot-start timing, confirming that the previous
8-10s figures were dominated by the kubelet readiness-probe interval rather than
actual wake latency.

| Path | Before (Kube upper bound) | After (DPC log) | Source of inflation |
|------|---------------------------|-----------------|---------------------|
| Hot  | 8s (also 3s when probe lands sooner) | **0.53-0.59s** (`t_wake`) | kubelet readiness probe delay (~10s period) |
| Warm | 53-62s (`T_actuation`) | **~50.5s** (`t_instance_create`) | requester scheduling + probe delay |
| Cold | n/a this run | -- | no cold-launcher iterations occurred in the validation run |

An independent cross-check from the vLLM launcher log on the same run reported a
wake transition of ~0.51s, consistent with the DPC-derived `t_wake`.

### Measurement conditions

- **Model:** `meta-llama/Llama-3.1-8B-Instruct`
- **FMA launcher image:** `ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/launcher:v0.6.0-alpha.13` (includes PR #522 timing fields)
- **vLLM:** 0.20.2, sleep mode enabled, `gpu_memory_utilization=0.95`, prefix caching off
- **Cluster:** OpenShift on IBM Spectrum Scale storage, NVIDIA GPUs
- **Run:** 10 iterations; hit rates this run were 30% hot / 70% warm / 0% cold

Numbers are representative of a single validation run; hot-start hit rate (and
[...]

PR #1583, opened 2026-07-01
Consume centralized DPC "HTTP call done" log line; align actuation baseline

### LINKED ISSUE #1547: Update DPC log parser to consume the centralized "HTTP call done" log line
**Component:** Harnesses

## Desired use case or feature

The FMA timing harness parses Dual-Pods Controller (DPC) logs to report per-request actuation latency (hot / warm / cold-launcher). This parser shipped to `main` in #1504. It derives those timings by reading a `httpCallStartTime` field off three specific DPC milestone log messages — the wake call, the instance-create call, and the readiness relay.

Upstream PR llm-d-incubation/llm-d-fast-model-actuation#586 ("Centralize HTTP call logging and metrics") changes how the DPC logs HTTP calls: instead of stamping a start time onto each individual milestone message, it emits a single centralized "HTTP call done" log line for every HTTP call, carrying the start time, a `purpose` discriminator, latency, and status. The per-message start-time fields the parser relies on are removed.

As a benchmark maintainer, I want the DPC log parser updated to consume this centralized log line, because once #586 (or its successor) lands, the parser on `main` will stop producing DPC-derived timings — silently falling back to the coarser Kubernetes-timestamp upper bounds, with the sub-second wake fidelity quietly going blank. This is a compatibility break against a feature that is already shipped, not a hypothetical one.

## Proposed solution

Update the parser to treat the centralized "HTTP call done" line as the source of HTTP-call timing, correlating each line to a request via the requester identity that the DPC already attaches to its log context, and distinguishing the actuation paths via the line's `purpose` value:

- the wake call provides the hot-start anchor,
- the instance-create call provides the warm-start anchor,
- the readiness-relay call provides the shared end time used by all paths.

The cold-launcher anchor is unaffected (it comes from a Kubernetes API call, not an HTTP call, and is not part of the centralization), so that path should continue to work unchanged.

Alongside the parser change: refresh the module's documented assumptions (the relevant log verbosity and the fact that timing now comes from the centralized line rather than per-message fields), and add test coverage that locks in the new contract — including that the requester identity is present on the centralized line, since that correlation key is the linchpin and is currently provided implicitly by the logging context.

## Alternatives

1. **Ask upstream to preserve the per-message start-time fields** (raised via review comments on PR #586). Rejected as a solution to this issue: even if accepted, it only delays the work — centralizing HTTP-call logging is the direction the DPC is moving, so the parser will eventually have to consume the centralized line regardless. At best this buys time; it does not remove the need for this change.
2. **Keep relying on Kubernetes-timestamp upper bounds only** (drop DPC-derived timing). Rejected: the whole value of the DPC parser is that it removes the multi-second inflation introduced by the kubelet readiness-probe interval; falling back loses sub-second wake fidelity.
3. **Pin the benchmark to a pre-#586 DPC build.** Rejected: not sustainable; blocks picking up other DPC improvements and only defers the work.

## Additional context

- Related upstream PR: llm-d-incubation/llm-d-fast-model-actuation#586 ("Centralize HTTP call logging and metrics").
- The parser this issue updates landed in #1504 (merged 2026-06-19), so the break would hit a feature already on `main`.
- Open dependency on the upstream contract: this work should land once the centralized line's fields (the requester identity and the `purpose` token values) are confirmed/guaranteed and the required log verbosity is documented upstream. The PR #586 review comments request exactly that, so this issue is the downstream half of that coordination — and ideally lands close to when #586 merges, to minimize the window where `main` reports degraded timing.
- Scope note: cold-launcher timing is out of scope here (unaffected by the centralization).

_(AI-assisted issue draft.)_

rubambiza:
## Summary

Updates the FMA DPC-log timing pipeline for two coordinated upstream changes in `llm-d-fast-model-actuation` (both present in its v0.6.0 release):

1. **Consume the centralized `"HTTP call done"` log line.** The dual-pods controller replaced its three per-message HTTP-timing log lines with one centralized line discriminated by a `purpose` field. The parser now keys on `purpose` tokens: `wake` → hot-start anchor, `create_instance` → warm-start anchor, `relay_ready` → shared readiness end (`relay_unready` ignored). Cold-launcher timing is unchanged (it comes from a Kubernetes API call, not an HTTP call). Without this, once the upstream change is picked up the parser silently falls back to coarse Kubernetes-timestamp upper bounds and the sub-second wake fidelity goes blank.

2. **Align the actuation baseline to container start.** The controller's actuation metric now measures from the requester `inference-server` container's `state.running.startedAt` rather than the requester pod's `creationTimestamp`. The harness's Kube-timestamp fallback is realigned to the same baseline so it agrees with the controller.

Closes #1547.

## Key change: explicit, non-silent timing source

The single `dpc_timing_available` boolean is replaced with a per-iteration `timing_source`. **The three values measure three different intervals (not one interval at three fidelities)** — each ends at requester readiness but subtracts from a different start point:

- `dpc` — measured inside the DPC log as `relay_readiness − httpCallStartTime` of the wake/create call. Baseline is the HTTP call start (after the container is up), so it is the tightest interval. Highest fidelity.
- `kube_container_start` — Kube fallback: `requester ready − container state.running.startedAt`. Matches the controller's own actuation-metric baseline; a coarser upper bound than `dpc` (starts earlier).
- `kube_pod_create` — Kube fallback reverted to `requester ready − pod creationTimestamp` because container-start was unavailable. Earliest baseline, coarsest, degraded.

A `kube_pod_create` reversion emits a `logger.warning` naming the requester (fired only for the *final* source, after DPC refinement, so a DPC-timed iteration never emits a spurious warning). Because the sources measure different intervals, per-iteration `timing_source` exists so consumers don't mix them blindly. It propagates through `native_to_br0_1.py` and renders as a three-way Source column in the analysis output; `dpc_timing_available` is retained as a derived property for backward-compatible readers.

## Testing

- Unit: `tests/test_dpc_log_parser.py` (15) + `tests/test_fma_functions.py` (18); full suite 562 passed, 31 skipped.
- Preserves the large-file streaming and partial-log regressions from #1504 (adapted to the new format; the streaming regression uses a non-indicator filler so the first indicator genuinely lands past 256KB).
- **Smoke test against a real controller log** (1.12 MB, from a Qwen3-8B run): the updated parser extracts 11 records with sub-second hot-start `t_hot` (0.536 / 0.537 / 0.542 / 0.591 s, median 0.539s) — the same log yielded zero records under the old parser. Corroborates the ~0.51s vLLM-log weight-reload figure.

## Draft status

Marked draft pending **cluster end-to-end validation**: confirm the emitted `result.yaml` shows `timing_source: dpc` with sub-second `t_wake`, and that any fallback iteration reports the correct source. The smoke test above exercised the DPC-log path against real data; the container-start fallback path is unit-tested only, and is what the cluster run will exercise. Will flip to ready for review once that passes.

### DIFF
--- a/llmdbenchmark/analysis/scripts/nop-analyze_results.py
+++ b/llmdbenchmark/analysis/scripts/nop-analyze_results.py
@@ -350,9 +350,24 @@ def write_fma_metrics(  # pylint: disable=too-many-locals,too-many-statements
                 else None
             )
             node = launcher_info.get("launcher_node", "")
-            source = (
-                "DPC" if launcher_info.get("dpc_timing_available", False) else "Kube"
-            )
+            # Three-way timing source so a degraded pod-create baseline is
+            # distinguishable from a container-start Kube fallback at a glance.
+            # Fall back to the legacy dpc_timing_available flag for older reports
+            # that predate the timing_source field.
+            source_map = {
+                "dpc": "DPC",
+                "kube_container_start": "Kube (container-start)",
+                "kube_pod_create": "Kube (pod-create)",
+            }
+            timing_source = launcher_info.get("timing_source")
+            if timing_source in source_map:
+                source = source_map[timing_source]
+            else:
+                source = (
+                    "DPC"
+                    if launcher_info.get("dpc_timing_available", False)
+                    else "Kube"
+                )
 
             pandas_datas.append(
                 {
--- a/llmdbenchmark/analysis/benchmark_report/native_to_br0_1.py
+++ b/llmdbenchmark/analysis/benchmark_report/native_to_br0_1.py
@@ -2543,6 +2543,10 @@ def import_nop(results_file: str) -> BenchmarkReportV01:
                     "units": Units.S,
                     "value": requester_info["dual_label_timestamp"],
                 }
+                ri["container_start_timestamp"] = {
+                    "units": Units.S,
+                    "value": requester_info.get("container_start_timestamp", 0.0),
+                }
                 info["requester_info"] = ri
 
                 info["actuation_condition"] = launcher_info["actuation_condition"]
@@ -2554,6 +2558,9 @@ def import_nop(results_file: str) -> BenchmarkReportV01:
                     "value": launcher_info.get("launcher_creation_timestamp", 0.0),
                 }
                 info["launcher_node"] = launcher_info.get("launcher_node", "")
+                info["timing_source"] = launcher_info.get(
+                    "timing_source", "kube_pod_create"
+                )
                 info["dpc_timing_available"] = launcher_info.get(
                     "dpc_timing_available", False
                 )
--- a/tests/test_dpc_log_parser.py
+++ b/tests/test_dpc_log_parser.py
@@ -96,9 +96,14 @@ class TestParseDpcLogFile:
         (flag dumps, leader election, reconcile churn) before the first
         relay/wake/create message, so the indicator can sit well past that boundary.
         """
+        # Use benign "HTTP call done" with a non-anchor purpose (e.g., get_health)
+        # as filler. This is an indicator message (so file is still recognized as DPC)
+        # but doesn't create any timing record (benign noise).
         filler_line = (
-            "I0603 15:00:00.000000  1 inference-server.go:145] "
-            '"Reconciling server" serverUID="noise" requesterName="noise-pod"\n'
+            'I0603 15:00:00.000000  1 inference-server.go:2074] "HTTP call done" '
+            'purpose="get_health" method="GET" url="http://10.0.0.1:8005/health" '
+            'requesterName="health-check" httpCallStartTime="2026-06-03T15:00:00.000000000Z" '
+            'latencySecs="0.001" statusCode="200"\n'
         )
         # Prepend well over 256KB of benign filler, then the real sample block.
         filler = filler_line * (300 * 1024 // len(filler_line) + 1)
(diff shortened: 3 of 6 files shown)


llmdbenchmark/analysis/scripts/nop-analyze_results.py
aavarghese, line 324: Should this time also start at container_start instead of create?

rubambiza:
## Cluster e2e validation — passed ✅

Ran standup → 10 iterations → teardown on FMA **v0.6.0** with **Qwen3-8B**, harness built from this branch tip (`f5fbc11`). Same FMA version / model / cluster as the earlier baseline run, which under the *old* parser produced `dpc_timing_available: false` on every iteration.

Emitted `result.yaml`:

- **`timing_source: dpc` on all 10 iterations** (derived `dpc_timing_available: true` consistent).
- **Sub-second hot-start `t_wake`:** 0.537, 0.537, 0.557, 0.587, 0.593, 0.597 s — the wake fidelity this change recovers, corroborating the ~0.51s vLLM-log weight-reload figure.
- **Warm `t_instance_create`** (DPC-internal HTTP timing): ~40.5s cluster + one 60.5s — far tighter than the Kube-timestamp upper bounds.
- No `403`/`forbidden`/`iterations: []`; run + teardown clean, GPUs freed.

### Scope note (honest coverage boundary)

Because every real actuation had DPC-log HTTP timing present, the run exercised only the **`dpc`** path. The Kube-timestamp fallback (`timing_source: kube_container_start` / `kube_pod_create`) did **not** fire on this healthy run — it triggers only when DPC-log timing is absent — so that path remains covered by unit tests (`tests/test_fma_functions.py`) rather than this live run.

Flipping to ready for review.

maugustosilva (approved):

PR #1589, opened 2026-07-02
fix: Some cleanup of CI flows for WVA+FMA so that WVA `make` targets are fixed

aavarghese:


config/scenarios/cicd/ocp-wva-fma-warmstart.yaml
```diff
@@ -0,0 +1,434 @@
...
+        baseUrl: https://thanos-querier.openshift-monitoring.svc.cluster.local
+        port: 9091
+
+      # -----------------------------------------------------------------------
+      # VariantAutoscaling — per-model scaling intent
+      #
+      # Should match the HPA min/max below; the VA caps what the controller
+      # is willing to compute, the HPA caps what actually gets applied.
+      # -----------------------------------------------------------------------
+      variantAutoscaling:
```
dumb0002, line 222: Are we creating VA objects? VA will be deprecated in the next WVA release - we should move to use only HPA objects. These can be addressed in a follow-up PR.

config/scenarios/cicd/ocp-wva.yaml
dumb0002, line 103: Same comment here: Are we creating VA objects? VA will be deprecated in the next WVA release - we should move to use only HPA objects. These can be addressed in a follow-up PR.

dumb0002 (approved):

PR #1590, opened 2026-07-02
FMA requester affinity got auto — resolver skips `acceleratorType` for disabled sections

aavarghese:


aavarghese:
@dumb0002 can you take a look if this is the right thing to do for FMA requester to be scheduled on a node to work with https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/jinja/24_fma-deployment.yaml.j2#L190-L210?

dumb0002 (approved):
Approving for testing the workflow `e2e`.

PR #1602, opened 2026-07-07
Fix WVA autoscaling for the FMA benchmark scenarios

dumb0002:
The WVA+FMA benchmark passes (warm-start and hot-start) never autoscaled — the HPA sat at <unknown> and stayed at 1 replica, making hot-start look no faster than baseline. Root cause: the WVA controller's ServiceAccount was unbound (tokenreviews/variantautoscalings forbidden), so it never published the `wva_desired_replicas metric`.

**Changes**

- `19_wva-kustomize.yaml.j2` — the upstream overlay pins the ClusterRoleBinding subjects to wva-system, and kustomize's namespace: doesn't rewrite an already-set subject. Added a patch to repoint each WVA-SA binding's subject namespace to the deploy namespace (leaving the prometheus/user-workload-monitoring binding alone). This is the actual fix for the autoscaling failure.
- `ocp-wva.yaml, ocp-wva-fma-warmstart.yaml, ocp-wva-fma-hotstart.yaml` — trimmed dead variantAutoscaling keys (enabled, maxReplicas, slo) left over from the deprecated VariantAutoscaling CR. Kept variantCost (still emitted as the llm-d.ai/variant-cost HPA annotation) and hot-start's minReplicas fallback. Comment-only/cleanup — no behavior change.

### DIFF
--- a/config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
+++ b/config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
@@ -157,6 +157,11 @@ scenario:
         minReplicas: 1
         variantCost: "10.0"
 
+      # Autoscaling — feeds the KEDA ScaledObject (28_wva-scaledobject.yaml.j2):
+      # min/maxReplicas -> min/maxReplicaCount, targetAvgValue -> the Prometheus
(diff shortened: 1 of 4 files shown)


dumb0002:
This PR is deprecated after WVA moved to KEDA (e.g., VA removal, etc)

PR #1613, opened 2026-07-08
Use Qwen3-8B for OCP nightlies and time FMA actuation from container start

aavarghese:


config/scenarios/cicd/ocp.yaml
```diff
@@ -2,18 +2,19 @@
...
     # Model
     model:
-      name: facebook/opt-125m
-      shortName: facebook-opt-125m
-      path: models/facebook/opt-125m
-      huggingfaceId: facebook/opt-125m
+      name: Qwen/Qwen3-8B
+      shortName: qwen-qwen3-8b
+      path: models/Qwen/Qwen3-8B
+      huggingfaceId: Qwen/Qwen3-8B
```
dumb0002, line 17: Should we use here the same model as in the WVA+FMA workflow?

aavarghese: WVA+FMA runs LARGE model Qwen/Qwen3-32B with heavy load taking time. Its a manually dispatched workflow as needed. 
This workflow FMA only runs nightly and needs to be quick and complete. 8B was good enough to give us usable actuation numbers. WDYT?

dumb0002: Ok, sounds good!

dumb0002 (approved):
LGTM

PR #1620, opened 2026-07-11
run: mount all workload profiles in debug mode

### LINKED ISSUE #1272: When running in "debug mode", not all workloads are made available on the harness pod
### Component

Run

### Describe the bug

Whenever debug mode (`-d`/`--debug`) is invoked during `run`, the following should happen: a) the harness `pod` should not be named after a particular workload, but with a generic name (as implemented in the old code) and b) all workloads, for all harnesses, should be available inside the `pod` (again replicating functionality from the pre-refactoring code).

### Steps to reproduce

Just `run` with `-d` option.

_No response_

hubunt:
## Summary

Fix run debug mode so the harness pod exposes workload profiles for all harnesses, not only the selected harness.

## Motivation

Fixes #1272.

When `llmdbenchmark run --debug` is used, the harness pod is intended to stay alive for manual inspection and debugging. The current path only renders, creates, and mounts the selected harness profile ConfigMap, so workloads from other harnesses are not available inside the debug pod.

## What changed

- In debug mode, render profiles for every built-in harness under `workload/profiles/*`.
- In debug mode, create one profiles ConfigMap per rendered harness.
- In debug mode, mount every profiles ConfigMap under `/workspace/profiles/<harness>`.
- Use a generic debug pod name prefix: `llmdbench-harness-debug-<suffix>`.
- Preserve non-debug behavior: only the selected harness profile ConfigMap is rendered, created, and mounted.
- Preserve local `--workload-file-path` debug behavior: an explicitly provided workload file is still used instead of expanding to all built-in profiles.
- Added regression tests for debug profile rendering, ConfigMap creation, template mounts, generic debug pod naming, and the current harness pod failure-query behavior.

## Testing

- `python -m pytest tests/test_render_profiles_workload_file_path.py tests/test_create_profile_configmap.py tests/test_harness_pod_template_scripts.py tests/test_deploy_harness_status.py tests/test_kube_helpers_pod_failures.py -q`
  - 24 passed.
- `python -m pytest tests -q`
  - 708 passed, 31 skipped.
- `pre-commit run`
  - Byte compilation, full unit tests, changed-scenario dry-run rendering, secret scanning, Ruff lint, and Ruff format passed.
- `git diff --check`

## Backward Compatibility

Non-debug run behavior is unchanged. Local `--workload-file-path` debug runs continue to render the explicitly provided workload file rather than expanding to all built-in profiles.


maugustosilva:
@hubunt this looks good, thank you for the contribution. Please sign your commits and fix the conflicts so we can merge.

hubunt:
@maugustosilva 
Resolved the rebase conflicts against the latest `llm-d/main` and pushed signed commit `48f988f`.

The two conflicted test files now retain both the upstream assertions and the debug-profile regression coverage. I also updated the pod-failure test fixture so its phase query returns a phase value and its final query returns JSON, matching the current helper implementation.

maugustosilva (approved):

PR #1628, opened 2026-07-12
fix: nop harness starts before launchers are ready for Requestors, adding GPU info to analysis table

aavarghese:
- [x] GPU UUID in the actuation report: captures the requester's assigned GPU from the `dual-pods.llm-d.ai/accelerators` annotation → adds a GPU column to `result.txt`. Lets you see per-iteration which GPU was used (warm/hot ↔ GPU).
- [x] Launcher-readiness barrier before the benchmark starts: waits for most launcher pods to be Ready so iteration 1 doesn't race launcher container startup (the "first-warm is 2×") instead of doing just `time.sleep(buffer_seconds)`

### DIFF
--- a/llmdbenchmark/analysis/scripts/nop-analyze_results.py
+++ b/llmdbenchmark/analysis/scripts/nop-analyze_results.py
@@ -374,6 +374,7 @@ def write_fma_metrics(  # pylint: disable=too-many-locals,too-many-statements
                     "Iteration": iteration["iteration"]["value"],
                     "vLLM Name": launcher_info["name"],
                     "Node": node,
+                    "GPU UUID": launcher_info["requester_info"].get("gpu_uuids", ""),
                     "Actuation Condition": actuation_condition,
                     "T_actuation(s)": ttrr,
                     "T_hot(s)": t_hot_val,
--- a/llmdbenchmark/run/steps/step_02a_fma_warmup.py
+++ b/llmdbenchmark/run/steps/step_02a_fma_warmup.py
@@ -1,8 +1,7 @@
 """Step 02a -- Wait for FMA launcher pool to warm up before benchmarking.
 
-Mirrors Braulio's pattern in llm-d-fast-model-actuation PR #579: after
-standup completes, deliberately wait for the fma-requester Deployment to
-become ``Available`` (i.e. at least one launcher is bound and its vLLM
+Wait for the fma-requester Deployment to become ``Available``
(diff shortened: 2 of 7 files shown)


llmdbenchmark/analysis/scripts/nop-analyze_results.py
```diff
@@ -374,6 +374,7 @@ def write_fma_metrics(  # pylint: disable=too-many-locals,too-many-statements
                     "Iteration": iteration["iteration"]["value"],
                     "vLLM Name": launcher_info["name"],
                     "Node": node,
+                    "GPU": launcher_info["requester_info"].get("gpu_uuids", ""),
```
dumb0002: Should we rename this column to `GPU UUID` instead?

llmdbenchmark/run/steps/step_02a_fma_warmup.py
```diff
@@ -115,15 +98,51 @@ def execute(
...
+        launcher_wait = cmd.wait_for_pods(
+            label=launcher_selector,
+            namespace=namespace,
+            timeout=timeout,
+            poll_interval=10,
+            description=f"FMA launchers ready (model={model_id_label})",
         )
+        if not launcher_wait.success:
+            context.logger.log_warning(
+                f"FMA warmup: launcher pods not all Ready within {timeout}s in "
```
dumb0002, line 123: Instead of just returning a `warning message`  - should we return a error or skip the next steps if not all launchers pods are ready?

dumb0002 (changes requested):

llmdbenchmark/run/steps/step_02a_fma_warmup.py
aavarghese: Since the populator will be creating so many launchers (one per GPU node), and failing if one of them has an issue seemed too risky to fail the night's whole run (especially if we weren't even going to use that specific launcher). If another launcher has an issue and we timeout, the DPC will eventually wait for it or not bind to it. So it will get dealt with it later on anyways.
This atlleast one launcher Readiness check is just a slightly better soft replacement for the blind time.sleep we had in there earlier. We dont want it to be a very strict check and wait for all and fail hard kinda condition. Just a best-effort optimization.

dumb0002: Ok, got it. Thanks for the detailed explanation.

dumb0002 (approved):

PR #1627, opened 2026-07-12
feat: enable eval-containers scenarios on OpenShift via --cluster-config

eladven:
## Summary

- **`--cluster-config` flag**: New CLI option (`--cc FILE`) that loads a user-maintained YAML file and deep-merges it as highest-priority overrides on top of `defaults.yaml → scenario.yaml`. Lets users inject cluster-specific values (storage class, service account, security context) without modifying scenario files.
- **Generic scenario files**: `eval-containers-aider-polyglot.yaml` and `eval-containers-gaia.yaml` are now cluster-agnostic (safe defaults: RWO, no storage class override). IBM VPC / anyuid-sa specifics moved to `config/cluster-configs/examples/openshift-ibm-vpc.yaml`.
- **Example cluster-config files**: Added `kind.yaml`, `openshift-generic.yaml`, and `openshift-ibm-vpc.yaml` under `config/cluster-configs/examples/` as starting-point templates.
- **OpenShift setup docs**: New `docs/openshift-setup.md` explains three-layer merge, one-time cluster-admin prerequisites (anyuid-sa ServiceAccount + SCC binding), and example run commands.
- **Fix: poll-based pod wait** (cherry-pick from Natan Ergas): replaced two-phase `kubectl wait` with a polling loop so the framework no longer hangs when short-lived agentic pods complete and disappear before the wait fires.
- **Fix: `dataAccess.runAsUser` conditional**: `06_pod_access_to_harness_data.yaml.j2` now only emits `runAsUser` when the value is set (non-null). Default changed to `null` so vanilla K8s clusters don't get an unexpected `runAsUser: 0`.
- **Fix: stale data-access pod**: `step_05` now detects the "pod updates may not change fields" immutability error and auto-deletes + recreates the pod instead of failing the run.

### kind
```bash
llmdbenchmark --spec examples/eval-containers-aider-polyglot run \
  --cluster-config config/cluster-configs/examples/kind.yaml \
  -p llmd-eval -m <model> -j 5
```

### OpenShift (generic)
```bash
# One-time cluster-admin setup (see docs/openshift-setup.md):
oc create serviceaccount anyuid-sa -n <namespace>
oc adm policy add-scc-to-user anyuid -z anyuid-sa -n <namespace>

# Run:
llmdbenchmark --spec examples/eval-containers-aider-polyglot run \
  --cluster-config config/cluster-configs/examples/openshift-generic.yaml \
  -p <namespace> -m <model> -j 5
```

## Test plan

[...]

PR #1630, opened 2026-07-13
feat: eval-containers on OpenShift via --cluster-config + GPU-served scenarios

natan-ergas:
## Summary

Enables the **eval-containers agentic benchmarks** (exgentic) to run on OpenShift
in two ways, and makes their scenarios cluster-portable:

1. **`--cluster-config` mechanism** — eval-containers scenarios are now
   cluster-generic; cluster-specific values (RWX storage class, service account,
   `runAsUser`) are injected at run time from a `--cluster-config` file instead of
   being hardcoded in the scenario. Example cluster-configs ship under
   `config/cluster-configs/examples/` (`kind`, `openshift-generic`,
   `openshift-ibm-vpc`, `openshift-pokprod`). The `-kind` scenario variants are
   removed in favor of this.

2. **GPU-served eval scenarios** — new `eval-containers-{gaia,aider-polyglot}-gpu`
   scenarios that **stand up a real vLLM model on GPUs** (the standard llm-d
   modelservice path) and drive the agents against that in-cluster endpoint,
   rather than run-only against an external LiteLLM URL. Plus a GPU nightly
   workflow (`reusable-ci-nightly-eval-gpu.yaml` + gaia/polyglot callers) modeled
   on the perf nightly (`standup → run → teardown`).

Bundled framework fixes: poll-based pod wait (avoids a hang when short-lived
agentic pods finish before `kubectl wait` fires), conditional
`dataAccess.runAsUser`, stale data-access-pod delete-and-recreate, namespace
existence-check before apply, and `harness.runAsUser`/`serviceAccount` added to
the config schema.

> Supersedes / continues the previously-closed #1627 (same fork→upstream
> contribution). Opened as a **draft** for review.

## Purpose

The eval-containers harness previously only ran run-only against IBM LiteLLM, and
its scenarios hardcoded IBM-cluster storage/SCC specifics — so they didn't port
to other clusters and couldn't exercise the standard llm-d "stand up a model and
serve it" path. This PR makes the scenarios portable (`--cluster-config`) and adds
a genuine GPU-served path so the agentic evals can run against a model llm-d
deploys, the same way the perf benchmarks do.

**Intended follow-up:** enter the **gaia** and **aider-polyglot** agentic runs
into the **nightly benchmark** (the GPU nightly workflow + callers in this PR are
the vehicle), so agentic quality is tracked nightly alongside the perf runs.

## How it was tested

- **`--cluster-config` / plumbing:** aider-polyglot ran end-to-end on an OpenShift
  namespace (5/5 pods completed, results collected, clean teardown) via
  `--cluster-config`; kind + OpenShift render/plan validated.
- **GPU-served path (pok-prod, 14× H100):** both scenarios run end-to-end,
  `standup → run → teardown`, serving **Qwen/Qwen3-32B**:
  - Model auto-discovered via the in-cluster epponly endpoint (no `-U`, no key).
  - **30-task runs** at `decode.replicas: 10` (10 GPUs, load-balanced): all 30
    harness pods complete, 30/30 in-pod gateways start, results collected +
    analyzed. gaia and aider-polyglot both exercised.
  - Scores are per-task `reward`/`passed` in each `task/result.json` (the harness
    does not compute an aggregate; pass-rate is tallied from the per-task files).
    Current scores reflect the served model / benchmark difficulty, not pipeline
    defects (gaia needs a working web-search tool; aider-polyglot is blind
    single-shot coding).
- **Render validation + detect-secrets pre-commit hooks pass** on the changed
  files. (pytest suite unaffected — the GPU changes are scenario YAML + one
  template line + workflow YAML, no Python.)

### Notes for reviewers
- Two model/agent lessons baked into the GPU scenarios: **serve Qwen3-32B**
  (gpt-oss-120b is incompatible with both agents — gemini-cli's JSON router can't
  parse its harmony tool calls; codex's `web_search` tool is rejected by
  vLLM-served gpt-oss), and **`decode.replicas` must track `-j`** (~2–3 tasks/GPU)
  or agents starve on the per-task timeout.
- `04_download_job.yaml.j2`: `pip install -U --user huggingface_hub` so the model
  download job has the `hf` CLI even on exgentic images (which ship
  `huggingface_hub 0.26.2`, pre-`hf`-CLI). Longer-term this belongs in the
  exgentic image build (`huggingface_hub>=0.34`).

Authored by @eladven (the `--cluster-config` feature, tenant fixes, docs) and
@natan-ergas (GPU-served scenarios + nightly). All commits DCO signed-off.


mengmeiye:
Hi @alonh @natan-ergas Could you please fix the CI errors? Thank you.

alonh:
```Hi @alonh @natan-ergas Could you please fix the CI errors? Thank you.```

Fixed. Thanks for the quick review!

File: llmdbenchmark/cli.py [hunk 1]
```diff
@@ -1869,6 +1805,17 @@ def cli() -> None:
         help="Manual OIDC token or API key for telemetry auth.",
     )
 
+    benchmark_parser.add_argument(
```
Vezio, line 1883: Where is `cluster_config` used within `llm-d-benchmark`? 

I see that - for example - `config/cluster-configs/examples/openshift-pokprod.yaml` is consumed as input to the parser, can't this information be stored the `scenario` itself? I believe we support these fields - or similar ones today - it would be a better way to consolidate and allow users to reproduce workloads by having it in one spot, rather than merged in under the covers. 

What is the benefit from separating it from the scenario and what options are supported there?

If it is _more usable_ to have it separated, we should pursue that - but I want to avoid many different ways to changes these knobs since that will become a maintenance burden. 

Thanks team!

Vezio: CC: @mengmeiye what are your thoughts here?

mengmeiye:
Hi @alonh @natan-ergas Thanks for the fix. The code looks good to me. Before we approve and merge it, could you please rebase your PR to upstream? Thanks again.

Vezio (changes requested):

llmdbenchmark/cli.py
eladven: The separation is intentional and important. The scenario file is committed to the repo and is meant to be portable/reproducible across clusters. Cluster-specific values — storage class names, service accounts, runAsUser — are environment-specific secrets/infrastructure facts that differ per cluster and per user. Committing them would either break portability (hardcoded ibm-spectrum-scale-fileset in an example others try to run) or require every user to fork and edit the scenario.

The --cluster-config is essentially the same pattern as values.yaml overrides in Helm — the base chart is generic, you layer your cluster-specific values on top without touching the chart.

The concrete problem it solves: before this mechanism, users had to manually patch 4-5 different fields across the scenario YAML just to run on a different cluster. Now they maintain one small file outside the repo.

alonh:
> Hi @alonh @natan-ergas Thanks for the fix. The code looks good to me. Before we approve and merge it, could you please rebase your PR to upstream? Thanks again.

Thanks @mengmeiye , rebased. Waiting to address the discussion by @Vezio and @eladven, happy to update according to your thoughts.

llmdbenchmark/cli.py
Vezio: Ok that is good then!

Vezio (approved):

Issue #1232, opened 2026-05-13
Inconsistency in use of `accelerator` vs `decode.accelerator`


kalantar:
I see an inconsistency. Is it important enough to fix?

I noticed in the template [13_ms-values.yaml.j2](https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/jinja/13_ms-values.yaml.j2), support for `decode.accelerator.{count, type, resource}`, I also see in [default.yaml](https://github.com/llm-d/llm-d-benchmark/blob/main/config/templates/values/defaults.yaml) that `accelerator` is also a top level field (a peer to `decode`). In the template, I see, for example, checks for `decode.accelerator.count`. If present, it is used, otherwise the template computes a value. However, I don’t see a check for `accelerator.count`. Should this be checked as well? ie, precedence `decode.accelerator` > `accelerator` > some template default behavior

Same for prefill, I imagine. 

I asked Claude to validate my observation. It says:

```
  For decode_accel_count (line 254):
  {% set decode_accel_count = (decode.accelerator.count |
  default(decode_default_gpus) | int) if decode.accelerator is defined else
  decode_default_gpus %}

  For decode_accel_resource (line 413):
  {% set decode_accel_resource = (decode.accelerator.resource if
  decode.accelerator is defined and decode.accelerator.resource is defined
  else accelerator.resource) %}

  The current precedence for count is:
  1. decode.accelerator.count (if decode.accelerator exists)
  2. Computed: decode.parallelism.tensor * decode.parallelism.dataLocal

  The current precedence for resource is:
  1. decode.accelerator.resource (if defined)
  2. accelerator.resource (top-level fallback) ✓

  So the resource field already respects the decode.accelerator > accelerator
  pattern. But count does not check accelerator.count as an intermediate
  fallback — it jumps straight from decode.accelerator.count to the computed
  default.
  ---
  Your observation is valid — there's an inconsistency. The resource field has
   the 3-tier pattern (decode.accelerator.resource > accelerator.resource),
  but count skips the middle tier. If someone sets accelerator.count at the
  top level (e.g., to force a non-standard GPU count for both decode and
  prefill without repeating it), it would be silently ignored.
```

It also, reasonably, observes:

```
  That said, accelerator.count doesn't exist in defaults.yaml today, so it's
  more of a design consistency / forward-compatibility question. 
```

Is it something that deserves a fix?

UgaTheDev:
[...]

PR #1638, opened 2026-07-14
fix: fall back to top-level accelerator.count for decode/prefill GPU …

elinacse:
## Why
#1232 
Several existing scenarios (cpu.yaml, kind.yaml, sim.yaml, eval-containers-*.yaml) work around the missing fallback today by repeating accelerator: { count: 0 } three times per file — once at the top level and once under each of decode/prefill — purely because there was no inheritance. This change makes the top-level value a real, usable default, closing the gap between it and accelerator.resource, which already worked this way.

## Changes
config/templates/jinja/13_ms-values.yaml.j2: added accelerator.count as an intermediate fallback in decode_default_gpus and prefill_default_gpus.
Backward compatibility
No behavior changes for any existing scenario: defaults.yaml doesn't set a top-level accelerator.count, and every scenario that currently relies on GPU sizing already sets it explicitly at the role level (tier 1, unaffected). The new tier only activates for configs that set a top-level accelerator.count without a role-level override — which no scenario in this repo does today.

### DIFF
--- a/config/templates/jinja/13_ms-values.yaml.j2
+++ b/config/templates/jinja/13_ms-values.yaml.j2
@@ -266,8 +266,8 @@ decode:
   priorityClassName: {{ decode_priority }}
 {% endif %}
 
-{# GPU count = accelerator.count if explicit, else tensor x dataLocal (each DP-local rank needs its own GPU) #}
-{% set decode_default_gpus = (decode.parallelism.tensor | int) * (decode.parallelism.dataLocal | int) %}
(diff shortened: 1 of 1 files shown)


elinacse:
@Vezio please review this PR . Thanks !

elinacse:
@mengmeiye please review the PR. Thanks !

mengmeiye (approved):
LGTM! Thanks for the PR.

PR #1639, opened 2026-07-14
Re-export missing KEDA standup helpers + trim inert autoscaling config

dumb0002:
**Summary**

Two related fixes to the Workload Variant Autoscaler (WVA) path:

1. _Bug fix_: WVA standup crashed at step 03 with module `llmdbenchmark.standup.wva` has no attribute `verify_keda_installed`.
2. _Cleanup_: the WVA scenario config exposed a set of variantAutoscaling / hpa knobs that are no longer consumed by rendering, with comments describing an obsolete (native HPA + prometheus-adapter) architecture. Trimmed to the fields that actually take effect and corrected the comments to reflect the current KEDA ScaledObject path.

**Changes**

`llmdbenchmark/standup/wva.py`
- Re-export verify_keda_installed and extract_prometheus_ca_cert from keda_prometheus_auth, matching what step_03 calls (and mirroring keda_saturation.py).

`WVA config (defaults.yaml + all WVA scenarios)`
- variantAutoscaling: keep only variantCost (the one live field → llm-d.ai/variant-cost annotation on the ScaledObject). Remove enabled, minReplicas, maxReplicas, and slo.{tpot,ttft} — no template or step reads them.
- hpa: remove dead fields but keep enabled (still read by smoketests/base.py to gate the autoscaling-aware replica-range check) and the live minReplicas / maxReplicas / targetAvgValue / behavior that drive the ScaledObject's minReplicaCount / maxReplicaCount / trigger threshold / advanced.horizontalPodAutoscalerConfig.behavior.
- Comments reworded: the block drives a KEDA ScaledObject (28_wva-scaledobject.yaml.j2), not a native HPA + prometheus-adapter. The hpa key name is retained for parity with the upstream WVA guide (KEDA also generates a real HPA under the hood). Also fixed stale references to a non-existent "template 27" and to a rendered VariantAutoscaling CR.

**Files touched:**
- `config/templates/values/defaults.yaml`
- `config/scenarios/guides/workload-autoscaling.yaml`
- `config/scenarios/cicd/ocp-wva.yaml`
- `config/scenarios/cicd/ocp-wva-fma-hotstart.yaml`
- `config/scenarios/cicd/ocp-wva-fma-warmstart.yaml`
- `config/scenarios/examples/multi-model-wva.yaml`

**Behavior impact**

- Standup: WVA-enabled scenarios get past step 03 (verified the module now resolves both attributes).
- Rendering: unchanged. Every removed field already had a | default(...) in 28_wva-scaledobject.yaml.j2, and only variantCost + the retained hpa.* fields were ever consumed. hpa.enabled is preserved so the smoketest's range check is unaffected.
- No VariantAutoscaling CR was rendered before or after; the removed SLO/replica knobs were already inert in this repo.

**Testing**

- Pre-commit passed byte-compile, unit tests (pytest tests/), and the changed-scenario dry-run render for all 5 WVA scenarios ([OK] each).
- Validated all 6 YAML files parse and that every WVA block now has variantAutoscaling == {variantCost} with hpa.enabled: true and the live hpa fields intact.

**Notes for reviewers**

- The hpa config key is intentionally not renamed to scaledObject. It maps 1:1 onto KEDA's horizontalPodAutoscalerConfig, KEDA generates a real HPA under the hood (which teardown/smoketests/state-capture correctly inspect via kubectl … hpa), and the name keeps parity with the upstream WVA guide. Renaming would be breaking for existing scenarios and DoE sweeps for a cosmetic gain.

### DIFF
--- a/config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
+++ b/config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
@@ -22,8 +22,9 @@
 # user-facing WVA well-lit path that the WVA make targets consume, use
 # guides/workload-autoscaling (WVA-only).
 #
-# All HPA / VariantAutoscaling knobs are listed below even when they match
-# defaults.yaml, so the full surface area is visible in one place.
+# The scaling knobs this scenario consumes (variantAutoscaling.variantCost +
+# the hpa.* block, which drive the KEDA ScaledObject) are listed below with
+# inline notes.
 #
 # Note: FMA owns the elastic vLLM tier here, so the modelservice decode and
 # prefill Deployments are explicitly disabled below (decode.enabled: false,
@@ -52,8 +53,8 @@
 #
 # How WVA tracks FMA: launcher pods carry `llm-d.ai/variant=<id>-fma` (also
 # applied at bind time via the ISC; label-based variant identification per
-# WVA repo PR #1145). Templates 27/28 retarget the VariantAutoscaling + HPA
-# at the FMA requester Deployment (`fma-requester-<id>`) when fma.enabled.
+# WVA repo PR #1145). Template 28 retargets the KEDA ScaledObject at the FMA
+# requester Deployment (`fma-requester-<id>`) when fma.enabled.
 #
 # Based on:
 #   https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.wva.md
@@ -129,6 +130,10 @@ scenario:
 
       namespace: ""
 
+      # Ref of llm-d/llm-d used for the WVA kustomize overlay (CRD + RBAC +
+      # controller manifests). Consumed by 19_wva-kustomize.yaml.j2.
+      repoRef: "main"
+
       image:
         repository: ghcr.io/llm-d/llm-d-workload-variant-autoscaler
         tag: v0.8.0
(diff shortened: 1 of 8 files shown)


config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
```diff
@@ -22,8 +22,9 @@
 # user-facing WVA well-lit path that the WVA make targets consume, use
 # guides/workload-autoscaling (WVA-only).
 #
-# All HPA / VariantAutoscaling knobs are listed below even when they match
-# defaults.yaml, so the full surface area is visible in one place.
+# The scaling knobs this scenario consumes (variantAutoscaling.variantCost +
+# the hpa.* block, which drive the KEDA ScaledObject) are listed below with
```
aavarghese, line 26: Maybe we need to remove https://github.com/llm-d/llm-d-benchmark/blob/main/llmdbenchmark/run/steps/step_02a_fma_warmup_hotstart.py#L115?

config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
```diff
@@ -129,6 +130,10 @@ scenario:
 
       namespace: ""
 
+      # Ref of llm-d/llm-d used for the WVA kustomize overlay (CRD + RBAC +
+      # controller manifests). Consumed by 19_wva-kustomize.yaml.j2.
+      repoRef: "main"
```
aavarghese, line 135: What does specifying `main` here do?

config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
dumb0002: Good catch. It's now fixed. Thanks!

config/scenarios/cicd/ocp-wva-fma-hotstart.yaml
dumb0002: I set it here mainly to make the ref visible/tunable per-scenario

aavarghese (approved):
LGTM!!

PR #1642, opened 2026-07-15
config: organize scenario stacks into method sections

### LINKED ISSUE #1064: Re-organize the scenarios `yaml` files with well-defined sections.
### Component

Setup/Standup

Currently, a scenario file contains multiple `stacks` (serving a single specific model), which each `stack` containing the following keys - name: `model`, `modelservice`,`fma`, `standalone`, `routing`, `inferenceExtension`,`prefill`, `decode`, `vllmCommon`, `storage`, `workDir`, `harness`, `images`. The main problem with the current arrangement is that it is not orthogonal. For instance, `inferenceExtension` applies **only** to `modelservice`, while `storage` or `vllmCommon` can be potentially applied to standup methods (`standalone`, `modelservice` and `fma`).

### Proposed solution

Re-organize the scenario file in order to make it more organized. The proposed design is `scenario` -> [`stack1`,...,`stackN`] and then within a stack the following sections: `standalone`, `modelservice`, `fma` and `common`. All `common` keys, such as `storage` and `model`, should be put under this section, and ALL "standup methods" should be able to inherit those keys. Additionally, "method-specific" keys (e.g., `prefill`) should be under the standup-method-specific key (in this case `modelservice`).

### Alternatives

_No response_

_No response_

hubunt:
## Summary

Reorganize scenario stacks into the four well-defined sections proposed in #1064: `common`, `standalone`, `modelservice`, and `fma`.

Common settings are expanded into the effective configuration inherited by every standup method, while method-specific modelservice settings can be nested under `modelservice`. The resolved `config.yaml` remains flat for existing templates and execution steps.

## Motivation

Fixes #1064.

Scenario stacks currently mix shared settings such as `model`, `storage`, and `vllmCommon` with method-specific settings such as `prefill`, `decode`, and `inferenceExtension`. This makes the scope of each field unclear and forces scenario files to rely on comment headers instead of an explicit YAML structure.

## What changed

- Added normalization for per-stack `common` sections before defaults, shared settings, and stack overrides are merged.
- Extended modelservice nesting to support `common`, `prefill`, `decode`, `inferenceExtension`, and `multinode` in addition to the existing gateway/router sections.
- Preserved override precedence: defaults < shared < stack < treatment/CLI.
- Added clear validation errors when `stack.common` or `shared.common` is not a mapping.
- Updated early workspace discovery to read `common.workDir`.
- Reorganized the FMA example and OCP CI scenario using the sectioned layout.
- Documented the new structure and backward-compatible flat spelling.
- Added regression coverage for common inheritance, method-specific normalization, multi-stack sibling metadata, invalid structures, and end-to-end scenario rendering.

## Testing

- `python -m pytest tests -q`
  - 674 passed, 31 skipped.
- `python -m pytest tests/test_scenario_sections.py tests/test_hoist_modelservice_sections.py tests/test_config_schema.py -q`
  - 62 passed.
- `pre-commit run`
  - Byte compilation, full unit tests, changed-scenario dry-run rendering, secret scan, Ruff lint, and Ruff format all passed.
- Rendered both migrated scenarios and compared their effective `config.yaml` dictionaries with the versions from upstream `main`.
  - `config/scenarios/examples/fma.yaml`: identical.
  - `config/scenarios/cicd/ocp.yaml`: identical.
- `git diff --check`

## Backward Compatibility

Legacy flat scenario keys remain supported and render unchanged. If both a legacy flat key and its sectioned spelling are present in the same scenario layer, the sectioned value wins. Treatment and CLI overrides continue to take final precedence.

The modelservice Helm chart's own `common` values can be written as `modelservice.common`, avoiding ambiguity with the stack-level `common` section.

## Checklist

- [x] My changes follow the style guidelines of this project.
- [x] I have performed a self-review of my changes.
- [x] I confirm that `pre-commit run` was run and all checks passed.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -93,15 +93,28 @@ Create a scenario YAML under `config/scenarios/` that overrides only the values 
 ```yaml
 scenario:
   - name: "my-deployment"
-    model:
-      name: Qwen/Qwen3-32B
-    decode:
-      replicas: 4
-    namespace:
-      name: my-namespace
+    common:
+      model:
+        name: Qwen/Qwen3-32B
+      namespace:
+        name: my-namespace
+    modelservice:
+      enabled: true
+      decode:
+        replicas: 4
+    standalone:
+      enabled: false
+    fma:
+      enabled: false
 ```
 
 Only the keys you specify are overridden. Everything else comes from `defaults.yaml`.
+The `common` section is inherited by every standup method. Method-specific
+settings belong under `standalone`, `modelservice`, or `fma`. Legacy flat
+scenario keys remain supported for backward compatibility.
+
+`modelservice.common` is distinct from the stack-level `common` section: it
+is passed through as the modelservice Helm chart's `common` values block.
 
 **Multi-stack scenarios - the `shared:` block.** The scenario file also
 accepts an optional top-level `shared:` key that's merged into every stack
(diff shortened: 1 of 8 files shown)


Vezio:
Pretty good intentions here to solve that issue, tagging as DNM for now until we can go through a bit more testing to ensure existing functionality doesn't break. Thank you for these changes, this is an amazing step forward to reduce redundant configurations.

Vezio:
In the meantime, can you resolve some of the conflicts being reported above?

hubunt:
Conflicts are resolved and the branch is rebased onto current main. The updated signed head is `04aeb0d`. I preserved the upstream OCP scenario additions (`maxModelLen`, `blockSize`, and vLLM flags) inside the new `common` section. Full tests pass: 674 passed, 31 skipped.

maugustosilva (approved):

PR #1645, opened 2026-07-15
step_07: optional gzip'd oc-exec result transfer (--fast-collect)

animeshtrivedi:
For some harnesses the results tree grows large enough that copying it with `oc cp` is painfully slow. This commit introduces a --fast-collect flag that changes `oc cp` to `oc exec | tar xz` adding streaming and compression.

cc:ed: @ruocco


llmdbenchmark/run/steps/step_07_deploy_harness.py
```diff
@@ -726,19 +727,89 @@ def _collect_treatment_results_discovery(
...
         )
 
+        # Opt-in via --fast-collect / LLMDBENCH_FAST_COLLECT. When off (the
+        # default) results are collected with the original ``oc cp`` path
+        # (slow: ~95 min/dir because the ~1.5 GB per_request_lifecycle_metrics.json
+        # tunnels through the apiserver exec stream at ~0.3 MB/s). The fast path
+        # copies the exact same files -- it only swaps ``oc cp`` for a gzip'd
+        # ``oc exec | tar`` stream, which crosses the tunnel far faster.
+        FAST_COLLECT = context.harness_fast_collect
+
```
Vezio, line 739: I'm wondering here, is there a more elegant solution to this? 

Right now the code is hard setting to `oc` but we have the following `cmd.kube` as well as some auto detect features - what if a user is merely using `kubectl` and doesn't have the `oc` binary installed? This kind of only applies then to those leveraging the `oc` binary, and more specifically targetted to OpenShift, where as we have other users running on vanilla `k8` using simply the `kubectl` binary.

There's also a series of subshells being opened here directly, which is ok, but just wondering if we can better harden this.

The intentions here are great and will be extremely beneficial!

Vezio (changes requested):

llmdbenchmark/run/steps/step_07_deploy_harness.py
animeshtrivedi: Hi @Vezio thank for your comment. Indeed this should be made independent, and not hard-coded on `oc`. Looking at the code around I am suspecting (i dont have a vanilla k8 cluster to test): 

```python 
kube_prefix = [cmd._kube_bin, *cmd._kubeconfig_args(), 
  "--namespace", namespace]
```

and then 
```python 
 inner = (
     " ".join(
          kube_prefix
         + [
             "exec",
             shlex.quote(data_pod),
...
```
^^ this should fix the `oc` nv `kubectl` issue. WDYT?

**On an elegant solution:** the thing I could think of is use `subprocess.Popen` instead of hand-constructing a shell command. But there as well, we may have to hand-code "--" commands. I have to look this up and validate how to construct this.

Vezio: I believe you can actually pass `CMD` to a method we have and it should format it for you so you don't need to assemble it...

```
       cmd = context.require_cmd()
       result = cmd.kube(
            "get",
            "crd",
            "-o",
            "json",
        )
```

For example is one way. This will appropriately append the `kube context` to the command as well, so you target the correct cluster. Please feel free to check this out in the code for more context, thanks!

Vezio: There's actually many examples in the file you're editing in `step_07_deploy_harness.py`

For example:

```
                    result = cmd.kube(
                        "apply",
                        "-f",
                        str(pod_yaml_path),
                        "--namespace",
                        harness_ns,
                        check=False,
                    )
```

Vezio: Then the only thing you'd be responsible would be _after the cluster operations_ i.e., the second half the pipe ` tar xz -C ` - I think there's something in python you can use that would handle this, maybe some standard library.

animeshtrivedi: working on it, thanks for the feedback @Vezio !

animeshtrivedi:
Here are some performance results that I did on an open-shift cluster (cc @Vezio ) 

On exgentic traces, just running a subset of 5 small traces.  
```yaml
load:
  type: trace_session_replay
  stages:
    - concurrent_sessions: 0
      session_rate: 1.0
      num_sessions: 5
  worker_max_concurrency: 10000
  base_seed: 42
  ...
  
  data:
  type: otel_trace_replay
  otel_trace_replay:
    skip_invalid_files: true
    use_static_model: true
    static_model_name: REPLACE_ENV_LLMDBENCH_DEPLOY_CURRENT_MODEL
    hf_dataset_path: "Exgentic/agent-llm-traces"
    include_errors: false    
    filter: "lambda x: x['benchmark'] == 'swebench' and x['max_tokens'] < 30000"
    bad_tool_call_handling: use_recorded
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: true
storage:
  local_storage:
    path: /workspace
```

| FAST_COLLECT | Data | Copy time | Speed |
|-------------|------|-----------|-------|
| false | 154MB | 126s | ~1.2 MB/s |
| true | 154MB | **3s** | **~51 MB/s** |

Naturally YMMV.

Vezio (approved):

Vezio:
Thanks for those changes - this should be good to merge once we get CICD validating it didn't break anything unintentionally. Thanks again!!

PR #1646, opened 2026-07-15
fix: vllm log parser needs updates to work with 0.24.0+ updates

aavarghese:


aavarghese:
Will get to this later, as needed. Closing for now

PR #1649, opened 2026-07-16
fix: rename shadowed log_plain overload to log_plain_console

### LINKED ISSUE #1197: Double occurence of function name
### Component

I don't know

### Describe the bug

Within llmdbenchmark/logging/logger.py:189:9 there are two functions that share the same name of `log_plain()`. I'm not sure if this was intentional, but the functions should ideally have separate names so that it is obvious when calling them which one is used. This is also throwing a lint error as below:

```bash

F811 Redefinition of unused `log_plain` from line 169
   --> llmdbenchmark/logging/logger.py:189:9
    |
187 |             handler.flush()
188 |
189 |     def log_plain(self, msg: str) -> None:
    |         ^^^^^^^^^ `log_plain` redefined here
190 |         """Write *msg* verbatim to every handler - no timestamp or level prefix.
    |
   ::: llmdbenchmark/logging/logger.py:169:9
    |
167 |         self.logger.info(self._apply_indent(msg), extra={"emoji": emoji} if emoji else {})
168 |
169 |     def log_plain(self, msg, emoji=None):
    |         --------- previous definition of `log_plain` here
170 |         """Log a plain message to console (no metadata) but full message to file."""
171 |         self.logger.info(self._apply_indent(msg), extra={"plain": True, "emoji": emoji})
```

For now I've ignored the lint error, but I'd like to know what steps we want to take to fix this and I can implement the fix. Thank you.

### Steps to reproduce

Go to llmdbenchmark/logging/logger.py:189:9.

_No response_

elinacse:
Two methods named log_plain() existed on LLMDBenchmarkLogger with different behavior; the console/file-split variant was always shadowed by the verbatim-write variant defined later, triggering F811 and leaving the first implementation dead code.

Co-Authored-By: Claude

fixes #1197

### DIFF
--- a/llmdbenchmark/logging/logger.py
+++ b/llmdbenchmark/logging/logger.py
@@ -170,7 +170,7 @@ class LLMDBenchmarkLogger:
             self._apply_indent(msg), extra={"emoji": emoji} if emoji else {}
         )
 
-    def log_plain(self, msg, emoji=None):
+    def log_plain_console(self, msg, emoji=None):
         """Log a plain message to console (no metadata) but full message to file."""
         self.logger.info(self._apply_indent(msg), extra={"plain": True, "emoji": emoji})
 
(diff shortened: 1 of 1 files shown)


elinacse:
@Vezio please review this, Thanks !

elinacse:
@mengmeiye please review the PR . Thanks !

mengmeiye (approved):
LGTM!

PR #1651, opened 2026-07-16
standup: derive and verify prerequisite CRDs

### LINKED ISSUE #1062: When installing kubernetes `gateway-api` and `gateway-api-inference-extensions`, automatically determine the list of `crds`
### Component

Setup/Standup

Currently, the list of `crds` for `gateway-api` and `gateway-api-inference-extensions` is hard-coded directly on the actuator which implements step 02 (i.e., "Install cluster-level admin prerequisites (CRDs, gateways, LWS, SCCs"). 

### Proposed solution

Based on the supplied version for `gateway-api` and `gateway-api-inference-extensions`, automatically derive a list of `crds` which are expected to be present on a cluster before deciding for the need for installation. This means that the currently existing hard-coded lists inside  `standup/steps/step_02_admin_prerequisites.py` will be removed.

### Alternatives

_No response_

_No response_

### LINKED ISSUE #1063: Check `crd` version before deciding to install a cluster-level admin pre-requisite
### Component

Setup/Standup

Currently, only the existence (or lack of thereof) of a list of `crd` is checked before decide to install a particular cluster-level "admin prerequisites".  

### Proposed solution

In addition to check existence, the actual version of each `crd` should be checked.

### Alternatives

_No response_

_No response_

hubunt:
## Summary

Derive the Gateway API and Gateway API Inference Extension CRD lists from their configured release manifests, and verify installed CRD release versions before skipping cluster-level prerequisites.

## Motivation

Fixes #1062.
Fixes #1063.

The admin prerequisite step previously used hard-coded CRD name lists and checked only whether those names existed. That allowed the lists to drift as upstream releases added or renamed CRDs, and it could not report when the cluster contained CRDs from a different release than the configured revision.

## What changed

- Render the configured Gateway API revision with `kubectl kustomize` and extract every `CustomResourceDefinition` name from the resulting manifest.
- Render the configured Gateway API Inference Extension release with `kubectl apply --dry-run=client -o yaml` and derive its expected CRD names.
- Support multi-document YAML and Kubernetes `List` manifests while deduplicating discovered CRD names.
- Remove the hard-coded Gateway API and inference-extension CRD lists, including the release-specific `.k8s.io` and `.x-k8s.io` variants.
- Query installed resources with `kubectl get crd -o json` and build a version-aware inventory from:
  - `*/bundle-version`
  - `app.kubernetes.io/version`
  - `helm.sh/chart`
- Normalize common release spellings and compare configured versions for Gateway API, Gateway API Inference Extension, agentgateway, Istio, and LeaderWorkerSet.
- If remote discovery is unavailable but CRDs from the relevant API group are installed, keep the existing cluster-scoped resources and warn instead of attempting a network reinstall.
- If Gateway API or inference-extension CRDs are present but their version differs, or the configured release contains additional CRDs, report the mismatch/missing names without overwriting cluster-managed resources.
- Install Gateway API or inference-extension CRDs automatically only when no CRDs from the relevant API group are present.
- Preserve existence-only compatibility when installed CRDs do not expose recognizable release metadata.
- Added regression coverage for manifest discovery, Kubernetes `List` resources, offline fallback, cluster-managed partial CRD sets, version extraction, and version matching.

## Testing

- `python -m pytest tests/test_admin_prerequisites_gateway_crds.py -q`
  - 14 passed.
- `python -m pytest tests -q`
  - 679 passed, 31 skipped.
- `pre-commit run`
  - Byte compilation, full unit tests, changed-scenario dry-run rendering, secret scanning, Ruff lint, and Ruff format passed.
- `git diff --check`

## Backward Compatibility

Externally managed gateways keep their existing skip behavior. CRDs without recognized version metadata continue to be treated as installed when all dynamically discovered names exist. Pre-provisioned or partially provisioned CRD groups remain owned by the cluster administrator: discovery failures, version mismatches, and newly expected CRDs produce warnings without mutating the existing cluster-scoped resources.

### DIFF
--- a/llmdbenchmark/standup/steps/step_02_admin_prerequisites.py
+++ b/llmdbenchmark/standup/steps/step_02_admin_prerequisites.py
@@ -1,6 +1,8 @@
 """Step 02 -- Install cluster-level admin prerequisites (CRDs, gateways, LWS, SCCs)."""
 
 from pathlib import Path
+from collections.abc import Mapping
+import re
 
 import yaml
 
@@ -13,32 +15,11 @@ from llmdbenchmark.executor.command import CommandExecutor
 # and is rendered at plan time.
 _AGENTGATEWAY_SCC_NAME = "llmdbench-agentgateway"
 
-GATEWAY_API_CRDS = [
-    "backendtlspolicies.gateway.networking.k8s.io",
-    "gatewayclasses.gateway.networking.k8s.io",
-    "gateways.gateway.networking.k8s.io",
-    "grpcroutes.gateway.networking.k8s.io",
-    "httproutes.gateway.networking.k8s.io",
-    "listenersets.gateway.networking.k8s.io",
-    "referencegrants.gateway.networking.k8s.io",
-    "tlsroutes.gateway.networking.k8s.io",
-]
-
-# Inference extension CRDs may use the graduated (.k8s.io) or
-# experimental (.x-k8s.io) API group depending on the installed version.
-# We check for both variants.
-GATEWAY_API_EXTENSION_CRDS_K8S = [
-    "inferencemodelrewrites.inference.networking.k8s.io",
-    "inferenceobjectives.inference.networking.k8s.io",
-    "inferencepoolimports.inference.networking.k8s.io",
-    "inferencepools.inference.networking.k8s.io",
-]
-GATEWAY_API_EXTENSION_CRDS_XK8S = [
-    "inferencemodelrewrites.inference.networking.x-k8s.io",
-    "inferenceobjectives.inference.networking.x-k8s.io",
-    "inferencepoolimports.inference.networking.x-k8s.io",
-    "inferencepools.inference.networking.x-k8s.io",
-]
+GATEWAY_API_GROUPS = ("gateway.networking.k8s.io",)
+GATEWAY_API_EXTENSION_GROUPS = (
+    "inference.networking.k8s.io",
+    "inference.networking.x-k8s.io",
+)
 
 AGENTGATEWAY_CRDS = [
     "agentgatewaybackends.agentgateway.dev",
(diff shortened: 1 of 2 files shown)


hubunt:
Updated this PR to cover both #1062 and #1063. Gateway API and inference-extension CRD names are now derived from the configured release manifests, while the installed inventory remains version-aware. New commit: `47212fd`. Validation: 12 focused tests passed; full suite: 677 passed, 31 skipped; all pre-commit hooks passed.

mengmeiye:
Thanks @hubunt for fixing issues in #1062 and #1063. 

Overall it looks good, but I have a few questions/concerns:

1. Discovery makes a mandatory network fetch on every gateway standup, before any skip decision — could this break pre-provisioned clusters?

`_discover_expected_crds` runs `kubectl kustomize/apply --dry-run` against the remote manifest URL unconditionally, before the skip check. Previously the skip was a pure local existence check, so a cluster with the CRDs pre-installed and no egress to `github.com/kubernetes-sigs/...` worked fine. Now, when discovery yields nothing, `expected_crds` is empty, both skip branches are guarded by `if expected_crds and …`, and it falls through to a network install that also needs that URL → standup fails.

Could we fall back to the prior existence-only behavior when discovery yields nothing — e.g. skip if the relevant CRD group (`*.gateway.networking.k8s.io`, `*.inference.networking.{k8s.io,x-k8s.io}`) is already present? Detecting newly-added upstream CRDs inherently needs the fetch, but "can't verify" probably shouldn't be treated as "must reinstall."

2. The version-mismatch / new-CRD path re-applies existing CRDs and hard-fails the standup on conflict — is this re-applying the right way?

`cmd.kube("apply", "--server-side", "-k", crd_url)` is now reached whenever a version doesn't match or a discovered CRD is missing — i.e. even when the CRDs already exist. On clusters where these cluster-scoped CRDs are owned by another field manager (ArgoCD, Flux, `kube-addon-manager` — common on OpenShift/GKE), server-side apply then returns a conflict and we append an error → standup fails. The test fixture in this PR would encode that exact `kube-addon-manager` conflict.

So maybe we should make the mismatch path warn-only, i.e., log a clear message that the installed CRD version differs from the configured revision and continue, rather than re-applying. These are shared, cluster-scoped resources — forcibly upgrading one (e.g. `inferencepools.inference.networking.k8s.io`) could change the schema out from under other workloads/controllers on the cluster, so for users who don't own the cluster probably shouldn't be mutating them on standup.

hubunt:
Thanks for the careful review. I reproduced both paths in the tests and updated the PR in `9e847fe`:

- If remote discovery fails and the cluster already has CRDs in the relevant Gateway API or inference-extension API group, standup now warns and leaves those resources unchanged. It no longer falls through to a second network-dependent install attempt.
- A known version mismatch is now warn-only for Gateway API and inference-extension CRDs. It does not run server-side apply.
- I applied the same ownership rule when a newly discovered release contains CRDs missing from an already-provisioned group: the warning lists the missing CRDs, but standup does not re-apply the manifest over resources managed by ArgoCD, Flux, kube-addon-manager, or another field manager.
- Automatic installation still runs when the relevant CRD group is completely absent.

Regression tests now cover offline discovery with pre-installed `.k8s.io`/`.x-k8s.io` groups, version mismatch without reinstallation, partial cluster-managed CRD sets, and installation when the group is absent.

Validation: 14 focused tests passed; full suite: 679 passed, 31 skipped; all pre-commit hooks passed.

mengmeiye (approved):
LGTM!

PR #1654, opened 2026-07-17
router: let epponly pick proxyType=agentgateway, not just envoy

### LINKED ISSUE #1653: epponly mode has no way to pick proxyType=agentgateway, always defaults to envoy
## what's missing

when gateway.className is epponly, the router values that get rendered only ever set up envoy sidecar stuff. there's no proxyType field in the templates anywhere, so the chart just falls back to its own default (envoy) no matter what you put in the scenario yaml.

looked at config/templates/jinja/12_router-values.yaml.j2, this is the only bit that touches router.proxy for epponly:

```
{% if _epponly
      and inferenceExtension.envoySidecar is defined
      and inferenceExtension.envoySidecar %}
  proxy:
{% if inferenceExtension.envoySidecar.args is defined ... %}
```

it only ever writes args/resources for envoy. grepped the whole repo for proxyType and got nothing, so it's not set anywhere else either.

## why this matters

the llm-d-router-standalone chart already has `router.proxy.proxyType: [envoy, agentgateway]` and it works fine on its own. but since llm-d-benchmark never actually sets that field, every epponly run is stuck on envoy. there's no way to benchmark agentgateway as the standalone router proxy through llm-d-benchmark right now.

## idea

add something like `inferenceExtension.proxyType` next to `envoySidecar`, default it to envoy so nothing breaks, and have the template pass it through as `router.proxy.proxyType` when epponly is active.

can try implementing this if that's useful

abhay1999:
fixes #1653

epponly (the standalone router topology) always deployed envoy as the proxy sidecar. theres no field anywhere in the templates to switch it to agentgateway, even though the chart itself already supports proxyType: [envoy, agentgateway] fine on its own.

## what changed

adds inferenceExtension.proxyType to the scenario schema, defaults to envoy so nothing existing changes. when epponly is active this gets passed through as router.proxy.proxyType.

also auto-sets epp.flags.secure-serving: "false" when agentgateway is picked and no explicit flags are set, since the chart requires that (agentgateway talks to EPP over plaintext grpc) and fails to render otherwise. if a scenario sets inferenceExtension.flags explicitly they need to add secure-serving: false themselves, documented that in standup.md.

## testing

- added tests/test_epponly_proxytype.py, 6 cases covering: nothing set keeps old behavior, agentgateway renders correctly, secure-serving gets auto set, explicit envoy still works, envoySidecar args/resources still work alongside proxyType, and non-epponly topologies ignore proxyType entirely
- ran the full existing suite, 297 passed 4 skipped, no regressions
- also verified through the real llmdbenchmark plan pipeline (not just my unit tests): built a scenario off cicd/kind.yaml with gateway.className: epponly + proxyType: agentgateway, ran plan, and confirmed the rendered 12_router-values.yaml has proxyType: agentgateway plus secure-serving: "false", while the existing router.proxy.resources values still flowed through correctly. ran the unmodified cicd/kind.yaml scenario the same way as a control, confirmed the output is byte identical to before this change (no proxy: block at all, since that scenario doesn't use epponly)

## docs

added a short section + example to docs/standup.md under the epponly part explaining the new proxyType field and the secure-serving requirement

### DIFF
--- a/config/templates/jinja/12_router-values.yaml.j2
+++ b/config/templates/jinja/12_router-values.yaml.j2
@@ -24,6 +24,9 @@
                                                   hack is gone)
      inferenceExtension.envoySidecar          -> router.proxy.{args,resources}
                                                  (standalone-chart Envoy)
+     inferenceExtension.proxyType             -> router.proxy.proxyType
+                                                 (epponly only, envoy [default]
+                                                 or agentgateway)
      inferenceExtension.extraServicePorts     -> router.extraServicePorts
      inferenceExtension.extraContainerPorts   -> router.epp.extraContainerPorts
      inferenceExtension.monitoring            -> router.monitoring
@@ -55,6 +58,7 @@
                     else images.inferenceScheduler %}
 {% set registry = _epp_image.repository.split('/')[0] %}
 {% set _epponly = (gateway.className | default('') == 'epponly') %}
+{% set _proxy_type = (inferenceExtension.proxyType | default('envoy')) | lower %}
 router:
 {# UDS tokenizer is a first-class chart feature on the llm-d-router
    charts. Toggling `inferenceExtension.sidecar.enabled` in a scenario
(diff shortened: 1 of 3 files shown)


abhay1999:
closing this, turns out its not needed anymore

went to rebase onto main after seeing the conflict and found PR #1541 already rewrote 12_router-values.yaml.j2 into a generic pass-through, scenarios can set router.proxy.proxyType: agentgateway directly now, no template mapping needed at all. so the core thing this PR was fixing is already fixed by that refactor.

also turned out my secure-serving auto-set logic here was solving a problem that doesnt exist, checked the chart's _validations.tpl again and it only fails if secure-serving is explicitly set to something other than false, leaving it unset (the default) is fine.

while double checking this with a real helm template run against the actual chart, did find one real bug thats still there: _normalize_router_block in render_plans.py hardcodes targetPort: 8081 for the epponly http service port, which is envoy's port. the chart requires targetPort to be http, 80, or omitted specifically when proxyType=agentgateway, so 8081 fails validation. going to open a separate, smaller PR just for that.

PR #1655, opened 2026-07-17
router: fix epponly http port for agentgateway proxyType

abhay1999:
## what

_normalize_router_block always hardcodes targetPort: 8081 for the epponly http service port, since thats where envoy's sidecar listens. but the chart's own validation rejects a raw port number there when proxyType=agentgateway, it only accepts targetPort omitted, 80, or the literal string "http". so any epponly scenario that sets proxyType: agentgateway fails at helm template time even though the scenario file itself looks totally fine.

fix just picks targetPort based on proxyType now, "http" for agentgateway, 8081 for envoy same as before.

## how i found and checked this

was originally trying to fix something else (no way to even pick proxyType=agentgateway for epponly), but while rebasing i noticed PR #1541 had already turned 12_router-values.yaml.j2 into a plain pass-through. so scenarios can just set router.proxy.proxyType: agentgateway directly now, didnt need a template change for that part at all.

wanted to make sure that path actually works end to end though, not just renders nice yaml. so i pulled the real llm-d-router-standalone chart and ran helm template against a rendered epponly + agentgateway scenario, and it blew up here:

```
Error: execution error at (llm-d-router-standalone/templates/epp.yaml:2:3): .Values.router.extraServicePorts[name=http].targetPort must be omitted, "80", or "http" when proxyType=agentgateway, got "8081"
```

patched _normalize_router_block locally, reran the exact same scenario through helm template again, passes clean this time (exit 0). also reran the default epponly scenario (no proxyType set) just to be safe, targetPort is still 8081 there, nothing changed for envoy.

## testing

- added 2 cases to the existing tests/test_normalize_router_block.py::TestEpponlyPort, one for envoy staying at 8081, one for agentgateway getting "http"
- ran the full suite after, 669 passed, 31 skipped, nothing broke

### DIFF
--- a/llmdbenchmark/parser/render_plans.py
+++ b/llmdbenchmark/parser/render_plans.py
@@ -1088,16 +1088,19 @@ class RenderPlans:
                 epp["flags"] = {"v": str(verbosity)}
 
         # --- 5. epponly: add HTTP service port for the in-pod proxy sidecar.
-        # envoy listens on 8081, agentgateway's chart validation only
-        # accepts targetPort omitted, 80, or "http" (not a raw port number).
+        # 8081 is envoy's listener port, the historical default here. Not
+        # every proxy wants that though -- e.g. agentgateway's chart
+        # validation only accepts targetPort omitted, 80, or "http", not a
+        # raw port number. Rather than special-casing proxy types, the
+        # target port is just a plain override: router.proxy.httpTargetPort.
         gw_class = (values.get("gateway") or {}).get("className", "")
         if gw_class == "epponly":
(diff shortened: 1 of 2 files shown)


abhay1999:
update based on feedback from nina on slack: dropped the proxyType check entirely, its now just router.proxy.httpTargetPort as a plain override, defaults to 8081 (same as before) if not set. a scenario going for agentgateway sets proxyType: agentgateway and httpTargetPort: "http" together, no branching on proxy type name in the code anymore. reran the same helm template check against the real chart, still passes clean

maugustosilva (approved):

PR #1660, opened 2026-07-22
Add opt-in hostPath-backed model storage with per-node download DaemonSet

### LINKED ISSUE #1433: feat: Add hostpath config to solve an issue with cold start
**What would you like to be added**: Hostpath config needs to be added to modelserver config on llm-d-benchmark project in both paths - WVA based autoscaling as well as EPP+KEDA based autoscaling. This feature then will be consumed by WVA make targets where we can see the benefits model load time and engine init time.

**Why is this needed**: Pod cold start is an issue for autoscaling across several dimensions. with this config we see that pod cold start reduced to 52 sec from 110 sec when we did initial bechmarking on IBM cloud.

aleskandro:
When `storage.hostPath.enabled: true` is set in a scenario, the benchmark creates a static PV backed by a node-local directory and binds the model PVC to it. A DaemonSet downloads the model to every target node (replacing the single download Job), so serving pods always find weights on local disk regardless of scheduling.

Resolves [#1433](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1433)

### DIFF
--- a/config/scenarios/guides/epp-keda-saturation.yaml
+++ b/config/scenarios/guides/epp-keda-saturation.yaml
@@ -248,6 +248,18 @@ scenario:
     storage:
       modelPvc:
         size: 1Ti
+      hostPath:
+        # TODO: should we default to enabled: true for WVA scenarios?
+        enabled: false
+        path: /mnt/models
+        storageClassName: hostpath
+        capacity: 1Ti
+        # Target the GPU nodes where decode pods will schedule.
+        # Replace with your cluster's actual node labels.
+        nodeSelector:
+          nvidia.com/gpu.present: "true"
+        tolerations: []
+        daemonSetTimeout: 3600
 
     # =========================================================================
     # WORKLOAD / HARNESS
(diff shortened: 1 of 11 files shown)


Vezio:
Looks pretty good, going to let CICD run

aleskandro:
Hi @Vezio , thanks for the review. I'm not sure the ruff-format errors are related to my changes, are they?

Vezio:
@aleskandro can you rebase? We just downgraded the `ruff` scanner, since the new version has way too many new rules it's checking for currently (not a bad thing...but we dont have bandwidth to verify they are non breaking).

Once you rebase we can merge most likely.

@maugustosilva and I are combing through this one and will have any comments soon, if any

aleskandro:
Hi @Vezio the rebase is done and tests pass :)

Vezio (dismissed):

Vezio:
Take a look at the Ruff format and markdown, I think there's been some reported issues. 

For `ruff` if you use the `precommit` that comes standard with `install.sh` you should be able to resolve those easily.

Otherwise this is great.

aleskandro:
Hi @Vezio , can we try again?

Vezio:
Rerunning workflows now!

Vezio (approved):

PR #1671, opened 2026-07-23
Adding FMA nightly in llm-d using well lit path

aavarghese:


llmdbenchmark/teardown/steps/step_01_uninstall_helm.py
```diff
@@ -138,25 +164,30 @@ def _delete_fma_crs(
...
-            "--selector=stood-up-via=fma",
-            "--namespace",
-            namespace,
-            "--ignore-not-found=true",
-            check=False,
-        )
+        # Delete the requester Deployment to start unbinding
+        dep_selectors = ["stood-up-via=fma"]
+        if guide_name:
+            dep_selectors.append(f"llm-d.ai/guide={guide_name}")
```
dumb0002: Here should be a replacement instead of an append? The requester pod will either have `stood-up-via=fma` or `llm-d.ai/guide={guide_name}`, is it?

config/scenarios/guides/fast-model-actuation.yaml
```diff
@@ -0,0 +1,98 @@
...
+      # Inline kustomize strategic merge patches applied on top of the
+      # guide's modelserver base. Default is empty -- the upstream guide
+      # already injects HF_TOKEN from the `llm-d-hf-token` Secret (which
+      # the benchmark's standup auto-creates from the HF_TOKEN env var),
+      # so no patch is needed for gated-model access.
+      #
+      # Use this list to override anything else upstream sets -- e.g.
+      # bump replicas or pin a priorityClassName:
+      #
+      #   patches:
```
dumb0002: A clarification question: what's the need for this patching block?

dumb0002 (changes requested):

llmdbenchmark/teardown/steps/step_01_uninstall_helm.py
aavarghese: Yes, nice catch!! It should be mutually exclusive. Thanks!

config/scenarios/guides/fast-model-actuation.yaml
aavarghese: Not needed - will remove. Thanks again!

dumb0002 (approved):
LGTM

PR #1692, opened 2026-07-29
Some more improvements to the FMA+KEDA/WVA workflows

aavarghese:
1. New reusable comparison-table script — `llmdbenchmark/analysis/scripts/fma_comparison_table.py`
- Single source of truth shared by both ci-benchmark-ocp-fma-keda and ci-benchmark-ocp-fma-wva,
- Reads the three pass dirs (baseline / FMA warm-start / FMA hot-start) and emits one comparison table with configurable column headers

2. `Run Info header` above the table — **served model, harness, workload profile, run duration, and total request count**

3. Enabled metrics collection for scale/replica info
- Added `monitoring.metricsScrapeEnabled: true` to all six scenarios (ocp-keda, ocp-keda-fma-{warm,hot}start, ocp-wva, ocp-wva-fma-{warm,hot}start)

4. Autoscaling metrics in the table
- Avg replicas, Max replicas, Avg KV-cache utilization (%), Avg queue depth (EPP), Avg pod startup (s), and Cost (avg replicas × GPU/hr)

5. Scenario tuning
- KEDA replicas/maxReplicas 10 → 6 to match the WVA arm (apples-to-apples pod counts)
- `boundLauncherTimeout: 1800` on the KEDA hot/warm-start scenarios for Qwen3-32B parallel load off shared storage

### DIFF
--- a/llmdbenchmark/analysis/scripts/fma_comparison_table.py
+++ b/llmdbenchmark/analysis/scripts/fma_comparison_table.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env python3
+"""Render the FMA autoscaling comparison table (baseline vs FMA warm/hot start).
+
+Shared by the EPP+KEDA (``ci-benchmark-ocp-fma-keda``) and WVA
+(``ci-benchmark-ocp-fma-wva``) nightly workflows
+
+Per-pass artifacts consumed:
+  * ``summary_lifecycle_metrics.json``              -- harness latency/throughput
+  * ``metrics/processed/replica_status.json``       -- aggregate_ready_replicas
+                                                       (avg/max replicas)
+  * ``metrics/processed/metrics_summary.json``      -- _aggregated.metrics EPP
+                                                       KV-cache / queue-depth means
+  * ``metrics/processed/pod_startup_times.json``    -- per-pod creation->Ready
+                                                       (Avg pod startup: baseline
+                                                       decode / FMA runtime requester)
+  (the metrics/* files are present only when monitoring.metricsScrapeEnabled).
+"""
+
+import argparse
+import glob
+import json
+import os
+
+
+# ---------------------------------------------------------------------------
+# Artifact location + loading
+# ---------------------------------------------------------------------------
+def newest_run_dir(root):
+    """Return the newest run root directory."""
+    roots = sorted(
+        os.path.dirname(d)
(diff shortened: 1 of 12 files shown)


llmdbenchmark/analysis/scripts/fma_comparison_table.py
```diff
@@ -0,0 +1,253 @@
...
+import json
+import os
+
+
+# ---------------------------------------------------------------------------
+# Artifact location + loading
+# ---------------------------------------------------------------------------
+def newest_run_dir(root):
+    """Return the newest run root directory."""
+    roots = sorted(
```
dumb0002: a `WARNING` from claude review: 

```
In fma_comparison_table.py:
roots = sorted(os.path.dirname(d) for d in glob.glob(os.path.join(root, "**", "results"), recursive=True) if os.path.isdir(d))
return roots[-1] if roots else ""
- It picks the "newest" run by lexical sort of the parent path, not by mtime. That's only correct if run directories sort lexically in chronological order (e.g. timestamped YYYYMMDD-HHMMSS names). If run dirs are named otherwise (or a non-timestamped dir sorts last), it silently picks the wrong run. A comment stating the naming assumption, or sorting by os.path.getmtime, would be safer.
- It matches any dir named results at any depth — if the artifact tree has nested/incidental results dirs, it could pick a wrong root. Low risk but worth a guard.
```

dumb0002:
@aavarghese, also posting the comments here: 

- The computation for the `Total requests`  is probably wrong because the `Inference-perf` uses **two workers** to generated the requests - see workload profile here:https://github.com/llm-d/llm-d-benchmark/blob/main/workload/profiles/inference-perf/shared_prefix_synthetic_heavy.yaml.in#L6. Hence `660` is the total request for a single worker only (`2 x 660 = 1320` ).

- It would be better to include only the `TTFT p99 (ms)` and the standard deviation - no need to include the `p50`, `p90`, `p95`.

llmdbenchmark/analysis/scripts/fma_comparison_table.py
aavarghese: mtime would be wrong here — these are GCS-downloaded artifacts, so file mtimes are the download time, not the run — so lexical sort of the timestamped names is the correct choice....left a comment

dumb0002 (approved):
LGTM

PR #1709, opened 2026-07-30
fix: scope nok8s container names, ports and workspace per stack

### LINKED ISSUE #1699: nok8s: Multiple nok8s stacks in one scenario collide on container names, host ports and the staged config directory
> `llm-d-planner v0.1.0`.

**Repro** - any scenario with two entries under `scenario:` that both set
`nok8s.enabled: true`:

```
llmdbenchmark --spec <spec-with-two-nok8s-stacks> --base-dir . standup
```

**Expected:** either per-stack isolation (stack-scoped container names and
`workspaceHostDir`), or standup refuses up front: "the nok8s method supports one
stack per host".

**Actual, dry-run:** both stacks emit byte-identical launch commands and both
report the same endpoint.

```
| [dry-run] would stage nok8s configs under /Users/.../.llmdbench/nok8s
| [dry-run] would stage nok8s configs under /Users/.../.llmdbench/nok8s
| ---> would have executed the command "docker run -d --name vllm-0 ... -p 8000:8000 ..."
| ---> would have executed the command "docker run -d --name vllm-0 ... -p 8000:8000 ..."
| ---> would have executed the command "docker run -d --name epp --network host ..."
| ---> would have executed the command "docker run -d --name epp --network host ..."
  Deployed Endpoints:
    nok8s-single: http://localhost:8081
    nok8s-second: http://localhost:8081
```

**Actual, live:** the two stacks race (`max_parallel_stacks` defaults to 4) and
one loses.

```
[06] Stack 'nok8s-second': FAILED: [06] nok8s_deploy (nok8s-second): FAILED - Failed to start container vllm-0: docker: Error response from daemon: Conflict. The container name "/vllm-0" is already in use by container "09aa868697ce...". You have to remove (or rename) that container to be able to reuse that name.
```

Worse than the visible conflict: `step_06_nok8s_deploy.py:85` runs
`docker rm -f <name>` for idempotency *before* each launch, so depending on
interleaving, stack B silently destroys stack A's already-running `epp`/`envoy`,
and the surviving containers serve stack B's config while standup reports stack
A's endpoint.

**Cause:** container names are hard-coded in
`config/templates/jinja/34_nok8s-containers.yaml.j2:21,35,44` (`vllm-{{ i }}`,
`epp`, `envoy`) with no stack qualifier, and `step_06_nok8s_deploy.py:55` and
`:117-128` stage every stack's EPP and Envoy config into the same
`nok8s.workspaceHostDir` (`~/.llmdbench/nok8s`).
`step_06_nok8s_teardown.py:48` hard-codes the same fallback names. Nothing
validates stack count.

UgaTheDev:
## Description

Fixes #1699.

Two nok8s stacks in one scenario resolved to the same container names, the same host
ports and the same staged config directory. The sharp edge is teardown: step 06's
`docker rm -f` matched by name, so tearing down one stack destroyed a sibling's
running containers. Ports and the workspace collided the same way.

Names, ports and the workspace directory are now scoped per stack. Per @mengmeiye's
design note on the issue, several stacks coexisting on one host is the intent and
should be supported, so this scopes rather than forbids.

Scoping alone is not enough, because two stacks can still resolve to the same suffix.
I reproduced both routes on the pre-fix branch: stacks named `chat one` and `chat-one`
slug to one suffix, and an explicitly shared `nok8s.nameSuffix` does the same. Both
rendered with no errors and produced an identical container set, so the destruction
path stayed reachable. `render_plans.py` now carries `_nok8s_name_claims` beside the
port claims, and `_validate_nok8s_host_claims` (renamed from `_validate_nok8s_host_ports`)
claims every resolved container name and raises a render error when another stack
already owns it.

I validated the resolved name rather than making `nok8s.nameSuffix` renderer-owned and
unsettable. Removing the override would not have closed the slug-collapse half of the
same hole, would silently ignore a key defaults.yaml documents as author-settable, and
would remove the only escape hatch for a stack name that reads badly as a container
suffix. Claiming the resolved name covers all three routes with one guard.

Four related defects fixed in the same pass, each reachable once stacks can coexist:

- The run phase synthesised one scenario-wide endpoint from `all_stacks_info[0]`, so a
  two-stack scenario benchmarked stack 1 twice and filed the results under each stack's
  name. `_nok8s_endpoint_url` now raises a `PhaseError` listing each stack and its port
  when there is more than one stack and no `--endpoint-url`, and `--stack` resolves that
  stack's own `listenPort`, so the documented workaround no longer needs `--endpoint-url`.
- The step 00 preflight used `_load_plan_config`, which sees only the first rendered
  stack, and never checked the new `envoy.adminPort`. A busy 19000/19100 became a 900s
  wait for an Envoy that never started. It now unions six ports across every stack and
  sums `replicas x tensorParallel` for the GPU-capacity warning.
- Accelerator sharing was undocumented and unguarded. At replicas 1 every stack emits
  `--gpus all`, so the second vLLM OOMs. docs/nok8s.md now shows per-stack
  `gpus: "device=N"`, and the preflight warns when more than one stack is left on the
  default all-devices selection.
- A non-numeric `nok8s.vllm.replicas` raised an uncaught ValueError out of the renderer,
  which was a regression against main. Replaced with the `isinstance(..., int)` guard
  used elsewhere.

Two smaller ones: intra-stack port duplicates passed silently (`envoy.listenPort` equal
to `vllm.hostPort` rendered clean, then Envoy failed to bind while readiness polled
vLLM twice and reported ready), so port claims are now keyed `(stack_name, config_key)`;
and a YAML-quoted port like `'8181'` disabled the guard for that key, so digit strings
are coerced before the isinstance check.

Not done here: having step 03 read each stack's own `34_nok8s-containers.yaml` so all
stacks can be benchmarked in one invocation. That is a feature change to the run
pipeline. This PR only has to stop the silent wrong-data path.

docs/nok8s.md now also tells a reader converting an existing single-stack scenario to
run teardown first, since the unsuffixed `vllm-0`/`epp`/`envoy` are no longer matched
by the new suffixed names and would keep holding their ports and device memory.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update (included)

## How Has This Been Tested?

Adds 5 tests to `tests/test_nok8s_plan.py` covering the container-name collision, the
shared `nameSuffix`, the same-stack duplicate port, the non-int replicas guard and the
run-endpoint guard.

```
$ python3 -m pytest tests/ -x -q -n2
749 passed, 31 skipped in 43.05s
```

Revert check, running the branch's test file against production code from main in a
throwaway worktree: `9 failed, 8 passed in 26.39s`. The two destruction-path tests fail
with `assert result.has_errors` -> False, which is main rendering both collisions
clean. The load-bearing pre-existing one fails verbatim as
`assert not ({'envoy', 'epp', 'vllm-0'} & {'envoy', 'epp', 'vllm-0'})`.

Exercised through the installed entry point, no GPU and no cluster:

- Slug-collapse two-stack scenario, `standup -n`: exit 1, three
  "nok8s container name 'X' is already used by stack 'chat one'" errors, and
  `grep -cE 'docker run|docker rm'` returns 0.
- Valid two-stack scenario: exit 0, names `{vllm-0,epp,envoy}-chat` and `-code`, every
  `docker rm -f` suffixed.
- `run -n` with no filter: exit 1 naming both stacks and ports. `run -n --stack code`:
  exit 0, three times `http://localhost:8181` (was 8081).
- Back-compat: the shipped single-stack `config/specification/guides/nok8s.yaml.j2`
  dry-run, standup docker commands and run endpoint, diffed against a main worktree is
  identical.

`pre-commit run`: py-compile, pytest, render-changed-scenarios, detect-secrets,
ruff-check and ruff-format all Passed. The SBOM hook failed and I did not "fix" it:
`python util/generate_sbom.py --check` on unmodified main prints the same
"SBOM out of date" here, so it is pre-existing host drift (no helm on this box,
different local .venv pins), not something this change introduces. The regenerated file
was reverted.

### Test Configuration

- Kubernetes version: none, and no GPU. A live standup was not run. "Two Envoys
  actually bind 19000 and 19100" and "the second vLLM OOMs on a shared GPU" are verified
  as emitted config and commands, not as running processes. I would rather flag that
  than imply otherwise.

## Checklist

- [x] My changes follows the style guidelines of this project
- [x] I have performed a self-review of my own changes
- [ ] I confirm that a full `./setup/standup.sh` -> `run.sh` -> `./setup/teardown.sh` sequence completed successfully
- [x] I confirm that `pre-commit run` was run and all checks passed
- [x] I have updated the documentation accordingly (`docs/nok8s.md`,
      `config/templates/values/defaults.yaml`)

No cluster and no GPU here. On the SBOM hook, see the note above.

### DIFF
--- a/config/templates/jinja/33_nok8s-envoy.yaml.j2
+++ b/config/templates/jinja/33_nok8s-envoy.yaml.j2
@@ -8,7 +8,7 @@ admin:
   address:
     socket_address:
       address: 127.0.0.1
-      port_value: 19000
+      port_value: {{ nok8s.envoy.adminPort }}
 
 static_resources:
   listeners:
--- a/config/templates/jinja/34_nok8s-containers.yaml.j2
+++ b/config/templates/jinja/34_nok8s-containers.yaml.j2
@@ -18,7 +18,7 @@ readiness:
   envoyPort: {{ nok8s.envoy.listenPort }}
 containers:
 {% for i in range(nok8s.vllm.replicas) %}
-  - name: vllm-{{ i }}
+  - name: vllm-{{ i }}{{ nok8s.nameSuffix }}
     kind: vllm
     image: "{{ nok8s.vllm.image }}:{{ nok8s.vllm.tag }}"
     hostPort: {{ nok8s.vllm.hostPort + i }}
@@ -32,7 +32,7 @@ containers:
     replicaIndex: {{ i }}
     extraArgs: [{% for a in nok8s.vllm.extraArgs %}"{{ a }}"{% if not loop.last %}, {% endif %}{% endfor %}]
 {% endfor %}
-  - name: epp
+  - name: epp{{ nok8s.nameSuffix }}
     kind: epp
     image: "{{ nok8s.epp.image }}:{{ nok8s.epp.tag }}"
     grpcPort: {{ nok8s.epp.grpcPort }}
@@ -41,7 +41,7 @@ containers:
     poolName: {{ nok8s.epp.poolName }}
     poolNamespace: {{ nok8s.epp.poolNamespace }}
     configMountDir: {{ nok8s.epp.configMountDir }}
-  - name: envoy
+  - name: envoy{{ nok8s.nameSuffix }}
     kind: envoy
     image: "{{ nok8s.envoy.image }}:{{ nok8s.envoy.tag }}"
     listenPort: {{ nok8s.envoy.listenPort }}
--- a/config/templates/values/defaults.yaml
+++ b/config/templates/values/defaults.yaml
@@ -1421,7 +1421,8 @@ nok8s:
   # Appended to every container name (vllm-N, epp, envoy). Filled in
   # automatically with "-<stack>" when a scenario has more than one stack, so
   # sibling stacks on one host don't share (and delete) each other's
-  # containers. An explicit value is preserved; single-stack scenarios keep
+  # containers. An explicit value is preserved, but two stacks resolving to
+  # the same container name is a render error; single-stack scenarios keep
   # the bare names.
   nameSuffix: ""
 
--- a/config/templates/values/defaults.yaml
+++ b/config/templates/values/defaults.yaml
@@ -1414,8 +1414,16 @@ nok8s:
   # can pull gated models from Hugging Face.
   hfTokenEnv: HUGGING_FACE_HUB_TOKEN
   # Host directory where rendered EPP/Envoy configs are staged and bind-mounted
-  # from (avoids sudo writes to /etc).
+  # from (avoids sudo writes to /etc). A scenario with more than one stack
+  # gets a per-stack subdirectory appended so stacks never overwrite each
+  # other's staged configs.
   workspaceHostDir: "~/.llmdbench/nok8s"
+  # Appended to every container name (vllm-N, epp, envoy). Filled in
+  # automatically with "-<stack>" when a scenario has more than one stack, so
+  # sibling stacks on one host don't share (and delete) each other's
+  # containers. An explicit value is preserved; single-stack scenarios keep
+  # the bare names.
+  nameSuffix: ""
 
   # vLLM inference worker(s)
   vllm:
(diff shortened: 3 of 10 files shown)


mengmeiye:
Hi @UgaTheDev Thank you for the PR! Can you please rebase your PR to upstream and we'll approve and merge it? Thank you.

mengmeiye (approved):

PR #1710, opened 2026-07-30
feat: CLI Overrides for Standup Configurations

Vezio:
## Summary

Adds `--set`, a CLI flag for overriding any scenario value at invocation time, so a variant that differs from an existing scenario in only a few fields no longer needs its own YAML file.

Previously the only ways to change a scenario value were to edit the file, keep a private copy, or write a `--cluster-config` file. Several shipped scenarios exist purely as near-duplicates of another one with two or three keys changed.

Multi-line values are folded onto one line. A value containing real newlines is read as a YAML plain scalar, so its line breaks collapse into spaces — which silently changes the meaning of a shell command

To keep the breaks, wrap the value in double quotes so \n is an escape: `--set 'decode.vllm.customCommand="export FOO=1\nvllm serve /model-cache/x"'`.

For a full multi-line customCommand, you should prefer the scenario file directly or --cluster-config.

--set is best suited to single-line values.

## What it does

`--set` takes dotted paths into the merged stack config, using the same paths and the same value semantics as the scenario YAML itself:

    llmdbenchmark --spec guides/optimized-baseline standup \
      -t kustomize --set kustomize.acceleratorBackend=gpu/sglang

Pairs are comma-separated and the flag is repeatable. Values are parsed as YAML, so numbers, booleans, lists and mappings mean what they would in the scenario file. It is available on every subcommand that renders templates
(`plan`, `standup`, `smoketest`, `run`, `teardown`, `experiment`), and can also be supplied via `LLMDBENCH_SET`.

Being available everywhere is deliberate: each lifecycle phase re-renders the plan, so a flag that existed only on `standup` would let the run phase render a config that disagrees with what was deployed.

## Multi-stack scoping

In a multi-stack scenario, prefixing a key with a stack name or an fnmatch glob scopes the override; unprefixed applies to every stack.

    --set 'qwen3-06b:decode.replicas=4,llama-31-8b:decode.replicas=1'
    --set 'wva.hpa.maxReplicas=6' --set 'llama-31-8b:wva.hpa.maxReplicas=2'
    --set '*-8b:decode.resources.limits.memory=64Gi'

When several selectors match a stack they apply by specificity, global, then globs, then exact names — so a general default plus a per-stack exception works regardless of the order the flags were typed.

This is orthogonal to `--stack`: `--stack` chooses which stacks are deployed, a selector chooses which stacks are modified. A selector matching no stack in the scenario is a hard error rather than a silent no-op.

## Precedence

    defaults.yaml → shared: → stack block → --cluster-config → --set
      → DoE setup.treatments → dedicated flags (-m, -t, --gateway-class, ...)

`--set` beats the stack's own block, but a DoE treatment beats `--set`, so a sweep is never flattened by a CLI flag. Dedicated flags remain authoritative because they are applied by resolvers that run after the whole merge — this
was pre-existing behavior that was previously undocumented.

## Naming

`--set` always means the scenario, on every subcommand. `run` and `experiment` keep `-o/--overrides` for workload-profile overrides, unchanged; the two are separate flags and can be combined in one invocation.

## Observability and secrets

Every applied override is logged with the value it replaced, so a plan rendered with overrides is auditable from the log alone. Credential-bearing paths are redacted in logs; the rendered config still carries the real value, since that is what produces the cluster Secret.

Overrides whose parent path does not exist warn as a probable typo, and values that YAML reads as something other than they appear (octal, hex, sexagesimal, dates) warn with the quoting escape hatch.

## Not in this PR

No scenario files are deleted. Collapsing the near-duplicate scenarios this flag makes redundant is left to a follow-up, because the candidates differ in more than the two keys they appear to — some inherit defaults the base guides
set explicitly, so parity needs decisions about intent rather than a mechanical swap.


maugustosilva:
Very good and welcome functionality, LGTM ASAP

maugustosilva (approved):

PR #1719, opened 2026-08-01
fix: nok8s readiness timeout dumps Envoy logs instead of the failing container

### LINKED ISSUE #1701: nok8s: On a vLLM readiness timeout, nok8s standup captures Envoy's logs instead of the failing vLLM worker's
> `llm-d-planner v0.1.0`.

**Repro** (the scout used a locally built image whose `vllm` entrypoint never
serves, to isolate this from lack of a GPU; the real-world trigger is the one the
docs name, "vLLM container exits at load"):

```
llmdbenchmark --spec <nok8s-spec> --base-dir . standup --nok8s-deploy-timeout 15
```

**Expected:** the failing container's logs are captured, since that is the only
place the cause (bad HF token, model too large for VRAM, wrong accelerator
image) is visible.

**Actual:** the error names the vLLM port, but the only artifact written is
`nok8s-envoy.log`, containing an unrelated Envoy startup warning:

```
[06] Stack 'nok8s-single': FAILED: [06] nok8s_deploy (nok8s-single): FAILED - Timed out waiting for http://localhost:8000/v1/models after 15s
```

```
$ cat ws9/.../setup/logs/nok8s-envoy.log
[2026-07-30 09:59:20.059][1][warning][main] [source/server/server.cc:959] There is no configured limit to the number of allowed active downstream connections. Configure a limit in `envoy.resource_monitors.global_downstream_max_connections` resource monitor.
```

No `nok8s-vllm-0.log` is produced.

**Cause:** `step_06_nok8s_deploy.py:99-104`

```python
ready_err = self._wait_ready(cmd, runtime, spec, context)
if ready_err:
    self._dump_logs(cmd, runtime, "envoy", context)
    return self._fail(stack_path, ready_err, [ready_err])
```

`"envoy"` is hard-coded, whereas `_wait_ready` (`:247-269`) already knows which
port timed out. `docs/nok8s.md:178-179` compensates by telling users to run
`docker logs vllm-0` by hand. `_dump_logs` already exists and is used correctly
for launch failures at `:92`.

**Suggested fix:** have `_wait_ready` return the failing container name and dump
that container's logs, or dump all of them on any readiness failure.

elinacse:
On a readiness timeout, step_06_nok8s_deploy hard-coded dumping Envoy's logs regardless of which container's port actually timed out -- _wait_ready checks each vLLM worker before Envoy, so the vLLM logs (the only place a bad HF token, OOM, or wrong accelerator image shows up) were never captured, only an unrelated Envoy startup warning.

Dump every launched container's logs on any readiness failure instead of guessing a single name.

Fixes #1701.

Co-Authored-By: Claude


UgaTheDev:
Thanks for taking this one. The "dump all of them" route is the right call: `_wait_ready` only returns an error string, so threading the failing port back out would have meant changing its signature for very little gain, and `_dump_logs` is already best-effort (`check=False`, `OSError` swallowed) so dumping every container cannot make the failure path worse.

I checked the test fails without the fix rather than only passing with it. Reverting the loop to the old `self._dump_logs(cmd, runtime, "envoy", context)` fails at `test_nok8s_plan.py:291` on the missing `nok8s-vllm-0.log`, so the regression is properly pinned. Whole file: 8 passed.

Two small things, neither blocking.

**1. The curl setup in the test is inert.** `ctx.nok8s_deploy_timeout = 0` makes `deadline = time.time() + 0`, so `while time.time() < deadline` never runs an iteration and curl is never invoked. I instrumented `_FakeCmd.execute` during the test to confirm: 0 curl calls, 2 `docker logs` calls (`vllm-0`, `envoy`). So `_FakeCmd(fail_substrings=("curl -fsS",))` and the comment above it describe something that does not happen. The readiness failure comes from the timeout being 0, not from curl failing.

That does not weaken what the test proves, since the dump loop is what you changed and it is exercised. It is just that a later reader would reasonably think curl-failure behaviour is covered here when it is not. Either dropping `fail_substrings` and saying the timeout is what forces the failure, or keeping it and adjusting the comment, would do.

For what it is worth, timeout=0 looks like the right choice regardless: `_wait_ready` sleeps 10 seconds between polls, so any non-zero timeout costs roughly 10s per port in the suite.

**2. `containers` vs `launched`.** The comment says "every launched container" but the loop iterates `containers`. They hold the same names at that point, since the `if errors: return self._fail(...)` above means every container started, so this is not a bug today. Iterating `launched` would just match the comment and stay correct if partial launch failures ever became non-fatal.

elinacse:
@UgaTheDev Thanks for the review , I have addressed your comments , Can you again review and approve the PR ?

maugustosilva:
Hello @elinacse thanks again for the PR. Please fix the conflicts and I will merge.

UgaTheDev:
@elinacse @maugustosilva heads up before the rebase, I think this one has been overtaken by #1757 and I want to flag it rather than have you spend time on the conflict.

#1757 added a rollback to the readiness-failure path, and that rollback dumps every launched container's logs before removing them, since `rm -f` destroys them:

```python
for name in reversed(launched):
    self._dump_logs(cmd, runtime, name, context)
    cmd.execute(f"{runtime} rm -f {name}", check=False)
```

On current main the readiness-failure path dumps Envoy, then calls `_rollback(launched)`, so `nok8s-vllm-0.log` is written today. I checked by applying the test from this PR on top of main with no other changes, and it passes:

```
tests/test_nok8s_plan.py::test_readiness_timeout_dumps_all_container_logs PASSED
1 passed, 64 deselected
```

That was not deliberate on my part, the overlap only became visible once both landed in the same function, and your diagnosis in #1701 was right.

Two things still worth salvaging if you want them: Envoy now gets dumped twice on that path, once explicitly and once via the rollback, which is harmless but redundant and could be a one-line deletion. And the regression test here is better than anything currently guarding that behaviour, so it would be worth keeping even if the source change is dropped. Happy to review either.

@maugustosilva up to you whether that is a trimmed version of this PR or just closing #1701.

mengmeiye:
Hi @elinacse Thank you for the PR! 
I do see some values in the function `test_readiness_timeout_dumps_all_container_logs`. Let's keep `tests/test_nok8s_plan.py` in this PR if we all agree?

UgaTheDev:
@mengmeiye agreed, and sorry for the slow reply — this went stale while waiting on me.

Keeping `tests/test_nok8s_plan.py` is the right call. To restate why in one line: #1757 fixed the behaviour incidentally, via a rollback that happens to dump every launched container's log before `rm -f` destroys them, so nothing currently pins the guarantee that a vLLM readiness timeout captures the vLLM worker's logs. If that rollback is ever reordered or made conditional, the regression comes back silently and no test notices. `test_readiness_timeout_dumps_all_container_logs` is what makes it explicit.

I re-confirmed it still passes on current main (`dfe7a42`) with only the test applied, no production change:

```
tests/test_nok8s_plan.py::test_readiness_timeout_dumps_all_container_logs PASSED
```

So @elinacse the conflict resolution should be small: drop the `step_06_nok8s_deploy.py` change, since main already dumps all containers, and keep the test. That leaves a test-only PR, which is the part with lasting value here.

mengmeiye:
Hi @elinacse thank you for updating the PR! Can you please resolve the DCO and ruff-format issues as checks are showing here? Thanks again!

UgaTheDev:
@elinacse both failures are small, and I reproduced them on your head (`ef9208a4`) so you do not have to hunt.

**ruff-format** is a single spot: `tests/test_nok8s_plan.py` needs two blank lines after the last assertion of `test_readiness_timeout_dumps_all_container_logs`, before the `# ----` banner comment that starts the next section. `ruff format --diff .` reports exactly this and nothing else, 1 file of 330:

```
@@ -655,6 +655,8 @@
     assert (logs_dir / "nok8s-envoy.log").exists()
+
+
 # ---------------------------------------------------------------------- #
 # Multiple nok8s stacks on one host (issue #1699)
```

So `ruff format tests/test_nok8s_plan.py` and commit is all it needs.

**DCO** is failing on one commit, `e177830d` ("fix: address PR review on nok8s readiness-timeout log dump"), which has no `Signed-off-by` trailer. Your other two commits are signed. Easiest fix without an interactive rebase:

```
git rebase --signoff dfe7a42
git push --force-with-lease
```

That adds the trailer to every commit on the branch and leaves the already-signed ones unchanged in content.

Worth noting the `Benchmark - Modelservice / agentgateway (Kind)` failure is almost certainly unrelated to a test-only change; I would not chase it until the other two are green and it has re-run.

PR #1730, opened 2026-08-04
Extract the Benchmark Report library into a standalone pip package (llmd-benchmark-report)

Bslabe123:
The Benchmark Report format is consumed by projects outside this repo, but until now the only ways to use it were to vendor the schema files or copy the directory around. This repo itself does the latter twice: the harness image installs it via a `_pyproject.toml` rename hack, and `llm_d_stack_discovery`'s image copies a slice of `llmdbenchmark` onto `PYTHONPATH`. kubernetes-sigs/inference-perf#461 is a third consumer that currently has to vendor the schema outright.

This PR makes the library a proper distribution, `llmd-benchmark-report`, living in `benchmark-report/` at the repo root. Canonical import name is `llmd_benchmark_report`; the bare `benchmark_report` name (used by the pod analysis scripts) and `llmdbenchmark.analysis.benchmark_report` both keep working as deprecated aliases with module identity preserved. Package versions track the schema version (`0.2.x` implements BR v0.2.x, additive-only within a minor line; policy in the README). Releases are cut by pushing a `br-v*` tag: the new workflow verifies the tag matches the package version, builds, smoke-tests, and publishes via PyPI trusted publishing.

While adding the schema drift test, it caught real drift: the committed `br_v0_1`/`br_v0_2` JSON Schemas were stale relative to the models (v0.2 was missing `session_performance` and the multimodal units). Both are regenerated here.

Until the first PyPI publish, `install.sh` (which all CI uses) installs the package from the checkout, so this PR is self-contained. Follow-ups after merge: register the PyPI trusted publisher, tag `br-v0.2.1`, then swap inference-perf's vendored copy for the dependency.

Marked WIP pending: PyPI project registration and owner sign-off from the report-format authors.


Bslabe123:
This is ready for review modulo two setup items that need admin or PyPI access. @clubanderson @petecheslock:

1. A `pypi` environment in repo settings. The publish job targets it, and adding required reviewers there would gate every publish behind a human approval.
2. A PyPI pending publisher for `llmd-benchmark-report` with owner `llm-d`, repository `llm-d-benchmark`, workflow `br-release.yaml`, environment `pypi`. Whoever registers it becomes the project owner on first publish, so if llm-d has a shared PyPI account that should be the one; otherwise I can register it and add you both as owners immediately after the first release.

Verification: the release workflow was exercised end to end on my fork by pushing a `br-v0.2.1` tag; the build job (tag/version guard, sdist+wheel, wheel smoke test) is green: https://github.com/Bslabe123/llm-d-benchmark/actions/runs/32765823662. The publish job is guarded to run only in llm-d/llm-d-benchmark, so fork tags act as dry runs. I also validated a report partial emitted by inference-perf (the first external producer) against the packaged models and JSON Schema; both pass, including across the additive observability fields added since inference-perf's vendored snapshot.

Vezio (approved):
This was a well needed, much improved, item. Thanks!

PR #1755, opened 2026-08-10
fix: nok8s preflight misses replica ports and Envoy admin, and passes silently without ss

### LINKED ISSUE #1703: nok8s: nok8s preflight port check misses replica ports and Envoy admin, and silently no-ops when ss is unavailable
> `llm-d-planner v0.1.0`.

`step_00_ensure_infra.py:131-137` builds the port list as:

```python
ports = [
    nok8s.get("vllm", {}).get("hostPort", 8000),
    nok8s.get("envoy", {}).get("listenPort", 8081),
    nok8s.get("epp", {}).get("grpcPort", 9002),
    nok8s.get("epp", {}).get("grpcHealthPort", 9003),
    nok8s.get("epp", {}).get("metricsPort", 9090),
]
```

Three gaps:

1. Only replica 0's port. With `nok8s.vllm.replicas: N` the workers occupy
   `hostPort .. hostPort+N-1` (`34_nok8s-containers.yaml.j2:24`), and
   `8001..800N` are never checked.
2. Envoy's admin port **19000** is hard-coded in `33_nok8s-envoy.yaml.j2` and
   listed as a required free port in `docs/nok8s.md:37`, but is not checked.
3. The check runs `ss -ltn` (`:220`) wrapped in `if res.success and res.stdout:`,
   so wherever iproute2 is absent it silently reports nothing. On the scout's
   host `ss` does not exist and the check passed with no output at all.
   `tests/test_nok8s_plan.py:157-170` only exercises the path where `ss`
   succeeds and reports no listeners, so this is untested.

**Suggested:** expand the port list, and log a note when the probe tool is
unavailable (`ss`, or `lsof -iTCP -sTCP:LISTEN` as fallback) instead of passing
silently.

---

UgaTheDev:
The nok8s preflight port check has three gaps, all of which let a standup proceed into a conflict it could have predicted.

`hostPort` is checked for replica 0 only. With `nok8s.vllm.replicas: N` the workers bind `hostPort .. hostPort+N-1` (`34_nok8s-containers.yaml.j2:24`), so `8001..800N` were never looked at. Envoy's admin port 19000 is hard-coded in `33_nok8s-envoy.yaml.j2` and listed as a required free port in `docs/nok8s.md:37`, but was not in the list either.

The third one is the reason I noticed. The check runs `ss -ltn` inside `if res.success and res.stdout:`, so on any host without iproute2 it reports nothing and passes. On macOS `ss` does not exist, so the check has never done anything there. `tests/test_nok8s_plan.py:157-170` only covered the path where `ss` succeeds and finds no listeners, so nothing caught it.

This expands the port list, adds `lsof -nP -iTCP -sTCP:LISTEN` as a fallback probe so the check works on macOS, and warns when no probe is available rather than passing silently. A warning that says it could not check is worth more than a silent pass that reads as all clear.

Four regression tests, each verified to fail before the change: every replica port checked, admin port checked, lsof fallback used when ss is absent, and the no-probe-available warning. Full suite matches the pre-existing baseline on main (same 4 unrelated failures in `test_config_schema.py`).

Fixes #1703

### DIFF
--- a/llmdbenchmark/standup/steps/step_00_ensure_infra.py
+++ b/llmdbenchmark/standup/steps/step_00_ensure_infra.py
@@ -15,6 +15,10 @@ from llmdbenchmark.executor.deps import (
 )
 from llmdbenchmark.utilities.cluster import print_phase_banner
 
+# Listening-socket probes, first one available wins. ss is iproute2 (Linux);
+# lsof covers macOS and hosts without iproute2.
+PORT_PROBES = ("ss -ltn", "lsof -nP -iTCP -sTCP:LISTEN")
+
 
 class EnsureInfraStep(Step):
     """Validate system dependencies and print cluster summary banner."""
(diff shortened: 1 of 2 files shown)


mengmeiye:
Hi @UgaTheDev The PR looks good to me! Can you resolve the conflicts so we'll approve and merge it? Thanks.

mengmeiye (approved):

PR #1756, opened 2026-08-10
fix: reject multi-stack scenarios on the nok8s method

### LINKED ISSUE #1699: nok8s: Multiple nok8s stacks in one scenario collide on container names, host ports and the staged config directory

UgaTheDev:
A scenario with two entries under `scenario:` that both set `nok8s.enabled: true` does not produce two stacks. It produces two attempts at the same one.

nok8s launches plain host containers with fixed names (`vllm-N`, `epp`, `envoy`), fixed host ports, and a single staged config directory (`nok8s.workspaceHostDir`). A second stack reuses all three. In dry-run both stacks emit byte-identical launch commands and report the same endpoint. Live, they race (`max_parallel_stacks` defaults to 4) and one loses to a name conflict.

The quiet failure is worse than the visible one: `step_06_nok8s_deploy.py:85` runs `docker rm -f <name>` for idempotency before each launch, so depending on interleaving, stack B destroys stack A's already-running `epp`/`envoy`, and the surviving containers serve stack B's config while standup reports stack A's endpoint. You get a green run measuring the wrong thing.

Issue #1699 offered two options: per-stack isolation, or refusing up front. This takes the second. There is nothing on this path to multiplex stacks over, and inventing stack-scoped names and workspace directories would be a large change to make a configuration work that has no reason to exist on a single host. Refusing during plan rendering costs one validation and tells the user exactly what to do instead.

Adds `docs/nok8s.md` note stating the one-stack-per-host constraint, since the docs did not say it.

One regression test, verified to fail before the change. Full suite matches the pre-existing baseline on main (same 4 unrelated failures in `test_config_schema.py`).

Fixes #1699

### DIFF
--- a/docs/nok8s.md
+++ b/docs/nok8s.md
@@ -22,6 +22,11 @@ equivalent of the upstream
 Use it for HPC/Slurm nodes, bare-metal boxes, or a single GPU workstation
 where standing up Kubernetes is not worth it.
 
+**One stack per host.** Container names, host ports and
+`nok8s.workspaceHostDir` are fixed, so a scenario with more than one stack on
+this method is rejected during plan rendering. Run one scenario per stack, or
+use a cluster-based method, to compare multiple stacks.
+
 ## Prerequisites
 
 `llmdbenchmark standup` validates these automatically in step 00; a missing or
--- a/llmdbenchmark/parser/render_plans.py
+++ b/llmdbenchmark/parser/render_plans.py
@@ -846,6 +846,30 @@ class RenderPlans:
         return errors
 
     @staticmethod
+    def _validate_nok8s_single_stack(
+        values: dict,
+        total_stacks: int,
+        stack_name: str,
+    ) -> list[str]:
+        """Reject multi-stack scenarios on the ``nok8s`` deploy method.
+
+        nok8s launches plain host containers with fixed names (``vllm-N``,
+        ``epp``, ``envoy``), fixed host ports and a single staged config
+        directory (``nok8s.workspaceHostDir``). A second nok8s stack on the
+        same host reuses all three, and standup's ``rm -f`` idempotency step
+        tears down the first stack's containers before launching its own.
+        There is nothing to multiplex the stacks over, so refuse up front
+        instead of letting them silently clobber each other.
+        """
+        if total_stacks <= 1 or not (values.get("nok8s") or {}).get("enabled"):
+            return []
+        return [
+            f"[{stack_name}] the nok8s method supports one stack per host "
+            "(container names, host ports and nok8s.workspaceHostDir are "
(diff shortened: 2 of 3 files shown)


mengmeiye:
Hi @UgaTheDev since this PR resolves the same issue as the PR #1709, can we safely close this PR (if you agree)? Thank you for your contribution!

UgaTheDev:
Agreed — this resolves the same issue as #1709. Closing as a duplicate. Thanks @mengmeiye!

PR #1757, opened 2026-08-10
fix: stop nok8s standup at the first container failure and roll back what launched

### LINKED ISSUE #1702: nok8s: A failed nok8s standup leaves containers running and keeps launching the rest of the stack after the first failure
> `llm-d-planner v0.1.0`.

**Repro**

```
# scenario with a bad nok8s.vllm.image
llmdbenchmark --spec <spec> --base-dir . standup
docker ps -a
```

**Expected:** stop after the first container fails to start (the stack cannot
work without it), and remove whatever was already launched before returning the
failure.

**Actual:** the run fails on `vllm-0` yet still pulls and starts `epp` and
`envoy`, which are left running on the host:

```
[06] Stack 'nok8s-single': FAILED: ... Failed to start container vllm-0: Unable to find image 'nonexistent-llmd-scout:v0.0.0-nope' locally
docker: Error response from daemon: pull access denied for nonexistent-llmd-scout, repository does not exist or may require 'docker login'
```

```
$ docker ps -a
envoy   envoyproxy/envoy:distroless-v1.33.2                Up 12 seconds
epp     ghcr.io/llm-d/llm-d-router-endpoint-picker:v0.9.0   Up 25 seconds
```

Same after a readiness timeout (`envoy`/`epp` `Up`, `vllm-0`/`vllm-1` `Exited`).
Because `epp` and `envoy` use `--network host`, they hold ports
9002/9003/9090/8081, so the next `standup` attempt trips the preflight's own
"Host ports already in use" warning.

**Cause:** `step_06_nok8s_deploy.py:82-97`. The `for c in containers:` loop
accumulates into `errors` and only returns after every container has been
attempted; there is no rollback of `launched` on failure. Recovery works
(`teardown --methods nok8s` cleans up, verified) but is undocumented as a
required step after a *failed* standup: `docs/nok8s.md:183` mentions it only
under "Ports busy".

Note the tension with issue 5: an auto-rollback must dump logs first, or it
destroys the only diagnostic evidence.

UgaTheDev:
A nok8s standup that fails on `vllm-0` still pulls and starts `epp` and `envoy`, and leaves them running on the host.

`step_06_nok8s_deploy.py:82-97` iterates every container, accumulating into `errors`, and only returns once all of them have been attempted. There is no rollback of what already launched. Because `epp` and `envoy` use `--network host`, they then hold 9002/9003/9090/8081, so the next standup attempt trips the preflight's own host-ports-in-use warning. Recovery works (`teardown --methods nok8s`), but the docs only mention it under "Ports busy", not as a required step after a failed standup.

This stops at the first launch failure and removes what was already started before returning.

The ordering matters and the issue calls it out: rollback must not destroy the evidence. Logs for launched containers are dumped before anything is removed, so the failure is still diagnosable afterward. `_dump_logs` is already best-effort (`check=False`, `OSError` swallowed), so the dump cannot make the failure path worse.

Two regression tests, both verified to fail before the change: one that the loop stops and rolls back, one that logs are captured before removal. Full suite matches the pre-existing baseline on main.

Note this touches `step_06_nok8s_deploy.py`, as does #1719 (readiness-timeout log dumping) and my #1756. The changes are in different functions, but whichever lands last will want a quick rebase.

Fixes #1702

### DIFF
--- a/llmdbenchmark/standup/steps/step_06_nok8s_deploy.py
+++ b/llmdbenchmark/standup/steps/step_06_nok8s_deploy.py
@@ -77,7 +77,6 @@ class NoK8sDeployStep(Step):
                 f"[dry-run] would stage nok8s configs under {workspace}"
             )
 
-        errors: list[str] = []
         launched: list[str] = []
         for c in containers:
             name = c["name"]
@@ -88,13 +87,15 @@ class NoK8sDeployStep(Step):
             )
             result = cmd.execute(run_cmd, check=False)
             if not result.success and not context.dry_run:
-                errors.append(f"Failed to start container {name}: {result.stderr}")
+                # The stack is unusable without every container, so stop at the
+                # first failure instead of launching the rest, and remove what
+                # is already up so it does not hold host ports (epp/envoy use
+                # --network host) against the next standup.
+                err = f"Failed to start container {name}: {result.stderr}"
                 self._dump_logs(cmd, runtime, name, context)
-            else:
-                launched.append(name)
-
-        if errors:
-            return self._fail(stack_path, "; ".join(errors), errors)
+                self._rollback(cmd, runtime, launched, context)
+                return self._fail(stack_path, err, [err])
+            launched.append(name)
 
         # Readiness: each vLLM worker, then Envoy.
         if not context.dry_run:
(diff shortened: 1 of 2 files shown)


mengmeiye:
Hi @UgaTheDev Thank you for the PR! Can you fix the ruff-format CI error and the conflicts? Thanks!

mengmeiye (approved):

Issue #1704, opened 2026-07-30
nok8s: the no-Kubernetes path renders the full Kubernetes manifest set and warns three times that helm/skopeo are missing


UgaTheDev:
> **Environment:** commit `2c1e326`; macOS 15 (Darwin 25.3.0) / Python 3.13.7 /
> Docker 29.1.5; no Kubernetes cluster, no GPU; helm, helmfile, yq, kind, podman
> absent. Reproduced in a venv with `pip install -e .` plus
> `llm-d-planner v0.1.0`.

Every nok8s invocation (`plan`, `standup`, `run`, `teardown`) emits:

```
Could not resolve WVA image tag: Unable to resolve latest tag for image "ghcr.io/llm-d/llm-d-workload-variant-autoscaler". Ensure skopeo, crane, or podman is installed and the image exists.
Could not resolve chart version for llmDInfra: Unable to resolve chart version for "llmDInfra". Ensure helm is installed and the repository is added, or provide a valid repo URL.
Could not resolve chart version for llmDModelservice: Unable to resolve chart version for "llmDModelservice". Ensure helm is installed and the repository is added, or provide a valid repo URL.
3 version(s) could not be resolved: wva.image.tag, chartVersions.llmDInfra, chartVersions.llmDModelservice. These will remain as 'auto' and must be resolved before deployment.
```

"must be resolved before deployment" is false on this path: nothing consumes
those chart versions, and `docs/nok8s.md:41-42` states helm/helmfile are not
required. A user following the docs is told they are missing tools they were just
told they do not need.

Relatedly, the render writes 36 files per nok8s stack, with real content in
Kubernetes-only manifests that will never be applied:
`05_namespace_sa_rbac_secret.yaml` (108 lines), `13_ms-values.yaml` (161),
`20_harness_pod.yaml` (117), `10_helmfile-main.yaml`, `04_download_job.yaml`,
`08_httproute.yaml`, `18_podmonitor.yaml`, `01/02_pvc_*.yaml`.
`render_plans.py:1664-1668` already computes `is_nok8s` to skip cluster-resource
resolution; the same flag could gate version resolution and the k8s template set.

---

UgaTheDev:
Half of this is fixed on current main (`dfe7a42`), so narrowing the issue to what remains rather than leaving it open at its original scope.

[...]

PR #1758, opened 2026-08-10
fix: skip Kubernetes-only version resolution on the nok8s path

UgaTheDev:
Every nok8s invocation (`plan`, `standup`, `run`, `teardown`) warns that skopeo/crane/podman and helm are missing, then says:

```
3 version(s) could not be resolved: wva.image.tag, chartVersions.llmDInfra, chartVersions.llmDModelservice. These will remain as 'auto' and must be resolved before deployment.
```

That last sentence is false on this path. Nothing consumes those chart versions when the deploy method is nok8s, and `docs/nok8s.md:41-42` states helm and helmfile are not required. A user following the docs is told they are missing tools the docs just told them they do not need, and that a deployment which is about to succeed cannot proceed.

`render_plans.py:1664-1668` already computes `is_nok8s` to skip cluster-resource resolution. This gates the WVA image tag and chart version resolution behind the same flag.

Scope note: issue #1704 also describes the render writing 36 files per stack, with real content in Kubernetes-only manifests that will never be applied. I left that out. Those files are reachable from teardown and from tooling I cannot rule out from the issue alone, and removing them is a much larger change with a real chance of breaking something that reads them. This PR fixes the false warnings only, which is the part that misleads users. The template-set question is worth a separate discussion.

Two regression tests, both verified to fail before the change. Full suite matches the pre-existing baseline on main.

Addresses #1704 (the version-resolution warnings; the template-set portion is left open)

### DIFF
--- a/llmdbenchmark/parser/render_plans.py
+++ b/llmdbenchmark/parser/render_plans.py
@@ -2033,7 +2033,9 @@ class RenderPlans:
 
         if self.version_resolver:
             try:
-                merged_values = self.version_resolver.resolve_all(merged_values)
+                merged_values = self.version_resolver.resolve_all(
+                    merged_values, skip_kubernetes=is_nok8s
+                )
             except Exception as e:
                 self.logger.log_warning(
                     f"Version resolution had issues for stack {stack_name}: {e}"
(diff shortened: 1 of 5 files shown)


mengmeiye:
Hi @UgaTheDev Thank you for the PR! Can you fix the ruff-format CI error and the conflicts so we can approve and merge it? Thanks!

mengmeiye (approved):

PR #1767, opened 2026-08-11
fix(eval-containers): bring nightly GPU eval to a scoreable baseline

natan-ergas:
## What this PR does

This PR fixes a set of independent bugs that together prevent the nightly GPU eval scenarios (`eval-containers-gaia-gpu` and `eval-containers-aider-polyglot-gpu`) from scoring above zero, or from producing a representative sample. Each fix is a separate commit; the commits that belong upstream are described below.

> **Note:** The last commit (`local: keep fork-only eval workarounds...`) contains three patches that are **deliberately not intended for upstream** (in-image runtime patches for grade.sh verbosity, codex web_search, and /logs collection). They are kept as a single identifiable commit so they are trivial to drop or cherry-pick separately.

---

### Fix 1 — EVAL_MODEL must be a bare handle `9df43bf` → `86c17a7`

`/opt/gateway/start` documents `EVAL_MODEL` as *"a BARE handle (e.g. aws/claude-opus-4-8)"* and renders it into a bifrost routing rule that carries the provider **separately**:

```sh
"targets": [{ "provider": "<wire>", "model": "<EVAL_MODEL>" }]
```

Upstream set `EVAL_MODEL="openai/${model}"`. The `openai/` prefix is not a provider selector — it becomes part of the model NAME. bifrost looks up `openai/Qwen/Qwen3.6-27B`, finds nothing, and returns HTTP 404 for **every** agent call. The failure is silent green: reward 0.0, llm_calls 0, harness exit 0. As it stands the nightly eval-GPU workflow cannot score above zero for this reason alone.

**Fix:** drop the prefix.

---

### Fix 2 — Raise the in-pod gateway request timeout `0e55c27` → `37d1f60`

bifrost's built-in per-request timeout is 30 s. Qwen3.6-27B emits a `<think>` block before answering and routinely exceeds it.

**Measured** (2026-07-30, two 20-task waves): 65/198 (33%) and 55/156 (35%) of gateway requests returned **504**, with durations pinned at 30002–30064 ms — a fixed deadline, not model variance. A successful call landed at 26 460 ms, 3.5 s under the wire. No task that saw 3+ 504s ever passed.

bifrost names the fix in its own error body (`default_request_timeout_in_seconds` in the `network_config`). The field is absent from the shipped template. `/opt/gateway/start` renders that template with plain `sed` and never parses the JSON, so an injected field passes straight through — no image rebuild needed.

The patch is keyed on `"allow_private_network": true` (appears exactly 3× in both images) and reverts from a saved copy if the substitution count is not exactly 3.

Default: **600 s**. Overridable via `EVAL_GATEWAY_TIMEOUT`.

---

### Fix 3 — `EVAL_TASK_LIST`: select an arbitrary task set `9034875` → `ee0d907`

`EVAL_TASK_OFFSET` can only express a contiguous range. The aider-polyglot dataset is **language-ordered** (cpp 0–25, go 26–64, java 65–111, javascript 112–160, python 161–194, rust 195–224), so every contiguous slice is single-language and not comparable to any other slice.

`EVAL_TASK_LIST` accepts a comma-separated list of task ids and maps each pod's 1-based position through it. Fully backward-compatible: the `else` branch is upstream's existing `EVAL_TASK_OFFSET` arithmetic, byte for byte.

---

### Fix 4 — Run the weight-download job on the python image `a020fc1` → `c4568e8`

The download job pip-installs `huggingface_hub` — Python and pip are all it needs — but it ran on `images.benchmark`, the multi-GB harness image. With `:latest` defaulting to `Always`, every standup re-pulled gigabytes that had nothing to do with fetching weights, causing the job to sit in ContainerCreating past its timeout. `images.python` (`python:3.10`) is already defined in `config/defaults.yaml`.

---

### Fix 5 — Serve the GPU eval scenarios at 32k with a reasoning parser `f0814f7` → `8ba1865`

Two changes, both required before either scenario can score honestly:

- **`maxModelLen: 32768`** (was 16000). An agentic task carries a repo's worth of context plus a `<think>` block; at 16k the request is rejected mid-solution.
- **`--reasoning-parser qwen3`**. Qwen3.6-27B is a hybrid-reasoning model. Without this flag vLLM returns the `<think>` trace as ordinary assistant content, so the agent receives reasoning prose where it expects a tool call.

Note: 32k halves per-pod concurrency — KV cache is a fixed token budget. Size `-j` off free GPUs, not off node capacity.

---

### Fix 6 — Set `EVAL_TIMEOUT` and `no-search` subset on GAIA `bb059da` → `f88ee8f`

The exgentic image's `run-agent` wraps each task in `timeout $EVAL_TIMEOUT`, defaulting to 300 s. At that ceiling the agent is killed mid-solution — the run silently under-reports the pass rate. `eval-containers-aider-polyglot-gpu` already carried `EVAL_TIMEOUT=1500`; the GAIA scenario was missing it.

Also set `EVAL_GAIA_SUBSET=no-search`, selecting the 54-task subset that drops tasks requiring a search engine. With no search backend wired up, those 111 tasks cannot pass and only depress the score.

---

### Fix 7 — Force the OpenAI wire for the gemini-cli polyglot harness `36af32b` → `365e048`

gemini-cli speaks the Gemini v1beta protocol, but llm-d/vLLM serves the OpenAI wire only. Without `EVAL_MODEL_API=openai`, the image routes gemini-cli requests to bifrost's gemini provider and fails with `no parser registered matching path suffix for: /v1beta/models/...:generateContent`. The agent then writes a stub solution (`???` was observed where a Rust type belonged) — indistinguishable from poor model quality.

`EVAL_MODEL_API=openai` selects the image's WIRE OVERRIDE mode: a single `cel_expression="true"` rule pinning `provider=openai`.

---

### Fix 8 — Collect and upload eval results + pod logs `8b382a6` → `314be58`

The nightly eval workflow had no result-publishing step: outputs lived only in the runner workspace and vanished. Adds workspace pinning and an `upload-artifact` step that runs before teardown and is `if: always()` so a failed run still produces diagnostics.

---

### GAIA (no-search subset, 54 tasks) — 2026-08-06

Model: **Qwen/Qwen3.6-27B** · maxModelLen: **32768** · 20 replicas (H100 80GB) · 3 waves

| Wave | Tasks | Passed | CTX32K rejections | HTTP 504 |
|------|-------|--------|-------------------|----------|
| Wave 0 | 20 | 10 | 2 | 4 |
| Wave 1 | 20 | 10 | 0 | 4 |
| Wave 2 | 14 | 7 | 0 | 0 |
| **Total** | **54** | **27 (50%)** | **2** | **8** |

The 16k baseline had 6–12 context rejections per run; 32k brought that to **2** (wave 0 only) then **0**. The web-search capability gap is a separate issue ([exgentic issue filed separately](docs/harness-workarounds-explained.md#8-codex-web_search----the-errors-you-saw)) and caps the score independently of these infrastructure fixes.

### aider-polyglot — 100-task cross-language run (2026-07-30 baseline, pre-fixes)

The 100-task polyglot run used `EVAL_TASK_LIST` to sample across all 6 languages (cpp/go/java/javascript/python/rust). Without it, every contiguous slice was effectively single-language. The `EVAL_TASK_LIST` feature in this PR is the enabler for that cross-language sampling.

---

## Checklist

- [x] Rebased on `main` (`765d2a3`)
- [x] Each fix is a separate commit with a standalone commit message
- [x] Local-only patches (grade.sh, web_search, /logs) isolated to final commit and **not** intended for merge
- [x] Scenario changes verified against `ghcr.io/exgentic/evals/gaia--codex-standalone` and `aider-polyglot--gemini-cli-standalone` (probed 2026-08-06, digests in docs/harness-workarounds-explained.md)

### DIFF
--- a/.github/workflows/ci-nightly-eval-gpu-ocp-aider-gemini.yaml
+++ b/.github/workflows/ci-nightly-eval-gpu-ocp-aider-gemini.yaml
@@ -1,6 +1,6 @@
 name: CI - Nightly Eval GPU on OCP (aider-polyglot / gemini-cli)
 
-# GPU-SERVED eval caller: stands up Qwen/Qwen3-32B on a single H100 via llm-d
+# GPU-SERVED eval caller: stands up Qwen/Qwen3.6-27B on a single H100 via llm-d
 # (modelservice) and drives the aider-polyglot / gemini-cli eval-containers
 # agent against that in-cluster served endpoint -- the standard llm-d path, NOT
 # run-only against LiteLLM. See reusable-ci-nightly-eval-gpu.yaml for the
--- a/.github/workflows/reusable-ci-nightly-eval-gpu.yaml
+++ b/.github/workflows/reusable-ci-nightly-eval-gpu.yaml
@@ -14,7 +14,7 @@ name: Reusable - Nightly Eval GPU (eval-containers, vLLM standup)
 # stood-up epponly endpoint. The in-pod bifrost gateway's OPENAI_API_KEY
 # defaults to a placeholder that llm-d ignores, so NO API key is forwarded and
 # NO `-U` is passed. EVAL_MODEL is derived from the scenario's model.name, which
-# must be the SERVED model id (e.g. Qwen/Qwen3-32B).
+# must be the SERVED model id (e.g. Qwen/Qwen3.6-27B).
 #
 # Target: OCP pok-prod (self-hosted runner + KUBECONFIG_DATA_OCP), same as the
 # run-only eval engine and the perf callers. Root for the eval image's /app
@@ -63,6 +63,11 @@ jobs:
       CLEANUP: ${{ inputs.cleanup }}
       # Forwarded to the model download job for gated models; harmless if empty.
       HF_TOKEN: ${{ secrets.LLMDBENCH_HF_TOKEN }}
+      # Pin the workspace so results land at a known path we can collect.
+      # Without --workspace the CLI falls back to the scenario's `workDir`
+      # (e.g. ~/data/eval-gaia-gpu) -- OUTSIDE the checkout, and on a
+      # self-hosted runner that path persists and mixes runs together.
+      WORKSPACE: ${{ github.workspace }}/eval-workspace
     steps:
       - uses: actions/checkout@v7
 
@@ -94,7 +99,7 @@ jobs:
           DRYFLAG=""; [ "$DRY" = "true" ] && DRYFLAG="--dry-run"
           llmdbenchmark --spec "$SPEC" $DRYFLAG standup \
             --cluster-config "$CC" \
-            -p "$NS" -t "$METHODS" --base-dir . \
+            -p "$NS" -t "$METHODS" --base-dir . --workspace "$WORKSPACE" \
             --pvc-bind-timeout "$PVC_TIMEOUT" \
             --modelservice-deploy-timeout "$MS_TIMEOUT"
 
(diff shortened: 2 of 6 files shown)


natan-ergas:
Closing in favour of a clean branch that excludes the local-only and CI-workflow commits. Reopening from natan-ergas:pr/eval-gpu-nightly-fixes.

PR #1768, opened 2026-08-11
fix(eval-containers): GPU-served agentic eval fixes + run-level metrics aggregation

natan-ergas:
Fixes for running the `eval-containers` agentic harness against an llm-d-served
model on GPUs, plus a run-level metrics aggregator. Every change below was found
by running the harness end to end on OpenShift (30 H100s, Qwen3.6-27B served by
llm-d) and is validated by the results at the bottom.

## What was broken

Each of these produced a *silent* failure — a run that reported success while
scoring ~0, or a metric that read as a real measurement while measuring nothing.

| Fix | Symptom before |
|---|---|
| `EVAL_MODEL` must be a bare handle | An `openai/` prefix became part of the model *name*; bifrost looked up `openai/<model>`, 404'd every agent call. reward `0.0`, empty agent stdout, harness rc `0`. |
| Raise the in-pod gateway timeout | bifrost's 30s per-request default is shorter than a reasoning model's generation. 33–35% of calls returned 504 pinned at 30002–30064 ms. No task with 3+ 504s ever passed. |
| `EVAL_TASK_LIST` | `EVAL_TASK_OFFSET` can only express a contiguous range, and aider-polyglot is language-ordered — so every slice was effectively single-language and scores were not comparable. |
| Weight download on `images.python` | The download job ran on the *harness* image. For agentic harnesses that is multi-GB with `:latest` ⇒ `imagePullPolicy: Always`, so every standup re-pulled gigabytes and the job timed out before transferring a byte. |
| 32k context + reasoning parser | At 16k, tasks carrying a repo's worth of context plus a `<think>` block hit the ceiling mid-solution. Without `--reasoning-parser` the `<think>` trace arrives as ordinary content and the tool call is lost inside it. |
| Force the OpenAI wire for gemini-cli | gemini-cli speaks Gemini v1beta, which bifrost cannot parse against an OpenAI-only upstream. Every call 400'd and the agent wrote a stub solution — a transport failure that reads as model quality. |
| Un-nest `images` from `common.storage` | Two scenarios indented `images:` one level too deep, so it parsed as `common.storage.images`, which nothing reads. The harness silently fell back to `ghcr.io/llm-d/llm-d-benchmark` — an image with no `/entrypoint.sh` — and the pod exited **127** in ~18s with no artifacts and nothing naming the image. |
| Read tokens from the provider span | The converter counted the HTTP span, but `gen_ai.usage.*` lives on the sibling `chat <model>` span. `total_token_rate` reported **0.0** and the token counters vanished from every report. Also fixed an `endswith` that double-counted reasoning tokens, and a span filter that matched only the OpenAI/Anthropic wires — so **gemini-cli reported 0 LLM calls and no latency at all**. |
| Retry the data-access pod lookup | The lookup returned `None` on the first failed API call, and the caller abandons collection. A one-second hiccup discarded two completed 30-task waves whose output was sitting intact on the PVC. |

## New: run-level aggregation

The framework runs one task per pod, so an N-task run left N single-task reports
and nothing that answered "how did the run do". `cross_treatment` cannot: it
reads each result dir as a separate experimental *treatment* and carries no
`observability` columns, so the score never reaches it.

`llmdbenchmark/analysis/aggregate_eval_containers.py` writes, under
`agentic-summary/` at the results root:

- **`agentic_run_report.yaml`** — a benchmark-report v0.2 for the whole run: pass
  rate, pooled LLM-call latency, token totals, task wall-clock, and per-language
  pass rates for `aider-polyglot`.
- **`agentic_tasks.csv`** — one row per task, units named in the columns.

Two metrics are reported conservatively rather than confidently:

- **time-to-first-token** is validated per call against its own call latency and
  the discarded count is published, because the raw attribute's unit is not
  dependable (one run implied 2,838 s inside an 823 s call).
- **task latency** comes from a harness-recorded wall-clock (`harness_start` /
  `harness_stop` / `harness_delta`, matching the other harnesses' key names), and
  is kept distinct from the span-derived `task_llm_span`, which omits agent
  setup, tool execution and grading.

## Validation

OpenShift, 30× H100, Qwen3.6-27B served by llm-d at `maxModelLen: 32768`,
`-j 30`, exercising the committed scenario values.

| | tasks | passed | LLM calls | tokens (in / out) | exit 124 | ctx rejections |
|---|---|---|---|---|---|---|
| **GAIA** (no-search, `codex`) | 54 / 54 | **27 (50.0%)** | 419 | 5.58M / 189k | 0 | 0 |
| **aider-polyglot** (`gemini-cli`) | 100 / 100 | **59 (59.0%)** | 1889 | 15.91M / 1.24M | 1 | 1 |

polyglot per-language: javascript 18/21 · python 16/19 · java 16/25 ·
rust 5/10 · cpp 4/11 · **go 0/14**.

**Timeout fix, measured:** 226 polyglot requests exceeded 30s and **224 returned
HTTP 200**. Against the unpatched 30s default every one of those would have been
killed. The 2 failures sit at 300,004 ms — a separate, 10× higher client-side
ceiling, not bifrost's. GAIA is silent on this fix (`slow30s=0` both waves), so
polyglot is its only evidence.

Some notes on the results:

- `go 0/14` is real: zero passes across fourteen tasks. cpp 4/11. Those two
  languages are where this model+agent combination fails due to performance
- Two GAIA tasks show `ServiceUnavailable` bursts confined to a 22-second window
  at the end of their wave: they were still calling when `run` scaled decode to
  0. Self-inflicted by wave teardown, not a model or config failure — but it
  scores as a failure, so it is worth knowing.

## Docs and examples

`docs/agentic_eval.md` now reflects the state of the fixes and new aggregation artifacts

The two eval containers example scenarios have been updated

## Test coverage

1253 passed, 32 skipped. new tests cover the span-selection and case-sensitivity
traps, the reasoning-token double-count, the `reward.txt` fallback, TTFT discarding,
schema validation of the emitted report, and the lookup retry.

### DIFF
--- a/config/templates/jinja/04_download_job.yaml.j2
+++ b/config/templates/jinja/04_download_job.yaml.j2
@@ -15,6 +15,22 @@ spec:
       containers:
         - name: downloader
           image: {{ images.benchmark.repository }}:{{ images.benchmark.tag }}
+          # Without this, a :latest tag makes Kubernetes default to
+          # imagePullPolicy: Always, so the node re-pulls the whole image on every
+          # standup even when it is already cached. That is invisible for a small
+          # image but not for a large one: with an agentic harness image (an agent
+          # CLI plus a multi-language toolchain, several GB) the pull outlived the
+          # download job's own timeout, so the job failed before transferring a
+          # single byte of weights -- surfacing as a weight-download timeout that
+          # points at storage or HuggingFace rather than at an image pull.
+          #
+          # IfNotPresent rather than changing which image this job runs: the pull
+          # policy is the actual cause, it is one line, and it keeps working
+          # whatever image a user points images.benchmark at. A user can also pin
+          # a hard version tag to get the same effect, but that is a workaround
+          # for a default that is wrong for this job -- nothing here needs a fresh
+          # pull, since the container only pip-installs huggingface_hub and exits.
+          imagePullPolicy: IfNotPresent
           command: ["/bin/sh", "-c"]
           args:
             - |
--- a/llmdbenchmark/utilities/kube_helpers.py
+++ b/llmdbenchmark/utilities/kube_helpers.py
@@ -33,7 +33,7 @@ DATA_ACCESS_LABEL = "role=llm-d-benchmark-data-access"
 # to what it guards: ~14s of polling against a wave of results that cost hours of
 # GPU time and cannot be regenerated once the harness pods are deleted.
 DATA_ACCESS_LOOKUP_ATTEMPTS = 5
-DATA_ACCESS_LOOKUP_DELAY_S = 3.0
+DATA_ACCESS_LOOKUP_DELAY_SECONDS = 3.0
 
 
 def _terminated_state_detail(prefix: str, state: dict) -> str:
@@ -105,7 +105,8 @@ def find_data_access_pod(
     cmd,
     namespace: str,
     attempts: int = DATA_ACCESS_LOOKUP_ATTEMPTS,
-    delay: float = DATA_ACCESS_LOOKUP_DELAY_S,
+    delay: float = DATA_ACCESS_LOOKUP_DELAY_SECONDS,
+    context: ExecutionContext | None = None,
 ) -> str | None:
     """Find the data-access pod by its well-known label.
 
@@ -123,7 +124,6 @@ def find_data_access_pod(
     afterwards with a plain ``kubectl cp``. Retrying costs a few seconds;
     not retrying costs the run.
     """
-    last_detail = ""
     for attempt in range(1, max(1, attempts) + 1):
         result = cmd.kube(
             "get",
@@ -142,13 +142,24 @@ def find_data_access_pod(
         # stdout is not proof of a pod name. Require a single bare token.
         if result.success and name and "\n" not in name and " " not in name:
             return name
-        last_detail = (
-            (result.stderr or "").strip() or name or "no pod matched the label"
-        )
+        detail = (result.stderr or "").strip() or name or "no pod matched the label"
         if attempt < attempts:
+            if context is not None:
+                # Log per attempt rather than only at the end: on a slow apiserver
+                # this is the only signal that collection is retrying rather than
+                # hung, and the reason often differs between attempts.
+                context.logger.log_warning(
+                    f"Data-access pod lookup attempt {attempt}/{attempts} failed "
+                    f"in {namespace} ({detail}); retrying in {delay}s"
+                )
             time.sleep(delay)
-    # Left to the caller to report: it knows which treatment was being collected
-    # and where the results still live on the PVC.
+        elif context is not None:
+            context.logger.log_warning(
+                f"Data-access pod lookup failed after {attempts} attempts in "
+                f"{namespace} ({detail})"
+            )
+    # The caller reports the user-facing failure: it knows which treatment was
+    # being collected and where the results still live on the PVC.
     return None
 
 
--- a/llmdbenchmark/utilities/kube_helpers.py
+++ b/llmdbenchmark/utilities/kube_helpers.py
@@ -29,6 +29,12 @@ CRASH_STATES = {
 
 DATA_ACCESS_LABEL = "role=llm-d-benchmark-data-access"
 
+# Retry budget for locating the data-access pod. Deliberately generous relative
+# to what it guards: ~14s of polling against a wave of results that cost hours of
+# GPU time and cannot be regenerated once the harness pods are deleted.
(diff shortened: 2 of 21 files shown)


config/templates/jinja/04_download_job.yaml.j2
```diff
@@ -14,7 +14,16 @@ spec:
...
+          # This job only needs Python + pip (it pip-installs huggingface_hub
+          # below), so it uses images.python rather than images.benchmark. With
+          # the benchmark image the weight download inherited whatever the
+          # harness image happens to be: for agentic harnesses (e.g. the
+          # exgentic eval images) that is a multi-GB image bundling an agent CLI
+          # and its toolchain, and since this template sets no imagePullPolicy a
+          # :latest tag defaults to Always -- so every standup re-pulled
+          # gigabytes of image unrelated to fetching weights, stalling the
+          # download job past its timeout before any bytes transferred.
+          image: {{ images.python.repository }}:{{ images.python.tag }}
```
Vezio: I think this is fine, but if the problem was simply `pulling` `latest` each time, a user could easily fix that by pinning to a _hard version_. If we ever update the `python` default image tag to `latest`, you will have to re-pull each time as well.

Although if the image size was the main issue, I could see that being an annoyance, but mostly a one time cost up front. 

The download container is mostly ephemeral anyway, so I think this is fine so long as it has all needed items for downloading models.

mengmeiye:
Hi @natan-ergas Thank you for the PR! Can you please sign all your commits? Thanks again!

llmdbenchmark/utilities/kube_helpers.py
```diff
@@ -95,24 +101,54 @@ def _pod_crash_details(pod: dict) -> list[str]:
...
+    namespace: str,
+    attempts: int = DATA_ACCESS_LOOKUP_ATTEMPTS,
+    delay: float = DATA_ACCESS_LOOKUP_DELAY_S,
+) -> str | None:
     """Find the data-access pod by its well-known label.
 
-    Returns the pod name, or ``None`` if not found.
+    Returns the pod name, or ``None`` if not found after ``attempts`` tries.
+
+    Retries because this lookup gates result collection, and a single failed
```
Vezio, line 115: Is it worth it to parameterize `retries` and `delay` within the `cli` for this type of action - or as an environment variable - so that these can be accessed beyond the script dynamically, rather than manually adjusted inline?

Not a needed requirement for the PR, but just a thought.

Vezio (dismissed):
Overall looks good to me, thanks for this @natan-ergas !

Please checkout the below comments - I will have CICD run in the meantime.

llmdbenchmark/utilities/kube_helpers.py
natan-ergas: Since this was a niche fix that solved some of my test runs, I added it as such without changing the cli, but that sounds like a better approach

config/templates/jinja/04_download_job.yaml.j2
natan-ergas: I will change this to have 'ifNotPresent'

Vezio:
Looks like a few CI have failed. Can you ensure the following:

- You are using `./install.sh` and leveraging `precommit`. That will ensure `ruff` format is correct. Otherwise, please `pip3 install ruff==0.15.11` and then run `python3 -m ruff format` and recommit the changes;

- Unit tests appeared to fail, but on deeper inspection seems transient based on a network issue on CI, I will rerun;

- It appears `markdown` is complaining about broken links in some documentation, please take a look at what applies to you, otherwise you may ignore and we can fix these later.

^ @natan-ergas please take a look ^

natan-ergas:
Addressed both review comments.

**Download job** — switched to `imagePullPolicy: IfNotPresent` and restored
`images.benchmark`, so the job no longer changes which image it runs. Net diff for
that file is now one line. Verified on OpenShift after a real teardown + standup:
the pod renders with `IfNotPresent`, runs the harness image, and reaches
`Succeeded` in 15s (it previously stalled in `ContainerCreating` past its timeout).

**Lookup retry** — parameterized as `--data-access-lookup-attempts` /
`--data-access-lookup-delay` (+ matching `LLMDBENCH_*` env vars), threaded through
`ExecutionContext` to all three call sites. Separate commit.

Also fixed the `ruff` failures: 3 files reformatted, and an `F841` that was a real
defect — a failure-reason variable I assigned but never read. The lookup now logs
each retry with its reason instead.

The `lint-and-test` failure is the markdown link checker on 17 files this PR does
not touch, all `Status: 0` on URLs that resolve fine (e.g.
`https://github.com/vllm-project/vllm.git`) — looks like CI rate-limiting.

PR #1775, opened 2026-08-11
Support EPP queue-based KEDA autoscaling in the ocp-keda scenarios

dumb0002:
### Summary

Adds EPP queue-size as a selectable KEDA scale trigger for the `ocp-keda`, `ocp-keda-fma-warmstart`, and `ocp-keda-fma-hotstart` scenarios, alongside the existing pool-saturation default — mirroring the upstream llm-d workload-autoscaling KEDA guides. No new scenario files; switching is a config override.

### Changes

- `config/templates/values/defaults.yaml` — added `llm_d_epp_flow_control_queue_size` to the metrics-scrape list (so the queue trigger has data and it's available to the comparison table); noted the default triggers mirror the upstream saturation guide.
- `workload/harnesses/process_metrics.py` — added `llm_d_epp_flow_control_queue_size` to the aggregation set + units.
- Scenarios (`ocp-keda*.yaml`) — documented, via a commented triggers: example, how to switch to EPP queue-based scaling (l`lm_d_epp_flow_control_queue_size, threshold 1`); removed the redundant pool-saturation example (it's already the default from `defaults.yaml`).

### Notes

- Saturation remains the default; queue is opt-in via the existing `eppKedaSaturation.scaledObject.triggers` override — no template change.

### DIFF
--- a/config/scenarios/cicd/ocp-keda-fma-hotstart.yaml
+++ b/config/scenarios/cicd/ocp-keda-fma-hotstart.yaml
@@ -155,19 +155,17 @@ scenario:
         maxReplicas: 6
         pollingInterval: 15
         unsafeSsl: true
-        # Defaults come from defaults.yaml; override the whole set here, e.g.
-        # (metricType + activationThreshold optional, default AverageValue/"0"):
+        # Triggers mirror the upstream llm-d workload-autoscaling KEDA guides.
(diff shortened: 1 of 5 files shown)


aavarghese (approved):

aavarghese:
Checking - is this trigger already in `fma_comparison_table.py`?

PR #1801, opened 2026-08-14
feat(cli): add cleanup subcommand to remove run leftovers from the namespace

### LINKED ISSUE #1789: `llmdbenchmark cleanup` subcommand to remove run leftovers from the namespace
### Component

Setup/Teardown

`llmdbenchmark run` tears down the harness launcher pod when a run finishes, but leaves the rest of what it created in the namespace:

- `pod/access-to-harness-data-workload-pvc`
- `service/llm-d-benchmark-harness`
- `pvc/workload-pvc`
- `configmap/llm-d-benchmark-preprocesses`
- `configmap/llm-d-benchmark-run-parameters`

Keeping the PVC between runs is useful (workload data survives), but there is no CLI path to remove any of this when the user is done. The only way out is a hand-written `kubectl delete` naming each resource. That causes three problems:

1. **The PVC bills until deleted.** `workload-pvc` is backed by shared storage (a Filestore instance on GKE). A user who follows a guide, runs one benchmark, and deletes their stack the way the guide shows still pays for the instance indefinitely, and nothing tells them it exists
2. **The resource names are CLI internals.** llm-d's `helpers/benchmark.md` will document the five-resource `kubectl delete` above so guides can link to it as a stopgap. That list silently breaks the next time the CLI renames anything.
3. **A stale PVC wedges retries.** A PVC's StorageClass is immutable and the CLI reuses an existing claim, so a run that failed against the wrong StorageClass fails every retry until the user discovers the claim and deletes it by hand (`ProvisioningFailed ... multi writer with mount access type` on RWO-only default classes is the common case).

### Proposed solution

```
llmdbenchmark cleanup --namespace <ns> [--keep-pvc]
```

- Deletes everything the CLI created in the namespace. Idempotent; missing resources are not an error.
- `--keep-pvc` preserves `workload-pvc` for users who want workload data to survive between runs (today's default behavior, made explicit).
- Implementation suggestion: label every resource the CLI creates (e.g. `app.kubernetes.io/managed-by=llmdbench`) and have `cleanup` delete by selector. That removes the name-list coupling for the CLI itself and lets external docs replace their hardcoded lists with one command.

Acceptance:

- After `run` followed by `cleanup`, `kubectl get all,pvc,configmap -n <ns>` shows none of the resources listed above.
- `cleanup` on a namespace with no benchmark resources exits 0.
- A run that failed on a wrong StorageClass can be retried after `cleanup` without manual PVC surgery.

### Alternatives

- Keep documenting the `kubectl delete` list in consuming repos (status quo). Rots whenever the CLI renames a resource, and every consumer maintains its own copy.
- Delete the PVC automatically at the end of each run. Removes the billing hazard but loses cross-run workload-data reuse, which is worth keeping as the default.

Motivated by llm-d/llm-d#2213, which moves the hand-maintained deletion list into `helpers/benchmark.md` ("Cleaning up harness resources") as a stopgap.

himanshu1573:
## What

Adds `llmdbenchmark cleanup --namespace <ns> [--keep-pvc]`, which removes everything a benchmark run leaves behind in the harness namespace:

- leftover harness launcher pods (`app=llmdbench-harness-launcher`)
- the data-access pod (`role=llm-d-benchmark-data-access`)
- `service/llm-d-benchmark-harness`
- ConfigMaps: `llm-d-benchmark-preprocesses`, `llm-d-benchmark-run-parameters`, `llm-d-benchmark-standup-parameters`, `llmdbench-harness-scripts`, and any `*-profiles` workload profile ConfigMaps
- `pvc/workload-pvc` (skipped with `--keep-pvc`; `--pvc-name` covers scenarios that override `storage.workloadPvc.name`)

## Why

Closes #1789. `run` tears down the launcher pod but leaves the rest in the namespace: the PVC bills against its storage backend until deleted, the resource names are CLI internals that consuming docs have to hardcode, and a stale claim wedges retries after a wrong-StorageClass failure.

## Design notes

- **Idempotent**: resources that no longer exist are skipped; a namespace with no benchmark resources (or a namespace that doesn't exist) exits 0, matching the acceptance criteria in the issue.
- **Pods are deleted before the PVC** so the `pvc-protection` finalizer can clear instead of leaving the claim stuck in Terminating.
- **No spec/plan required**: unlike the other subcommands, `cleanup` only needs `--namespace` (env: `LLMDBENCH_NAMESPACE`) — it deletes by name/label, so it works after the user has deleted their scenario or stack. It is dispatched like `results`, before plan rendering.
- The deletion set lives in one place (`llmdbenchmark/interface/cleanup.py`), so llm-d's `helpers/benchmark.md` can replace its hand-maintained `kubectl delete` list with this command. The issue also floats labeling every CLI-created resource (`app.kubernetes.io/managed-by`) and deleting purely by selector; happy to follow up with that if preferred, but name/label-based deletion keeps this PR small and works for resources created by older CLI versions too.

## Testing

- New unit tests in `tests/test_cleanup_command.py` (full deletion set, `--keep-pvc`, empty-namespace no-op, pod-before-PVC ordering, custom PVC name) following the existing fake-executor style.
- Full suite: 1141 passed, 32 skipped.
- `ruff check` / `ruff format --check` (0.15.11, as CI runs) clean.
- Manual `--dry-run` exercises the exact kubectl commands end-to-end.

Docs: added a section to `docs/lifecycle.md` after teardown, and listed `cleanup.py` in the README package layout.

### DIFF
--- a/llmdbenchmark/cli.py
+++ b/llmdbenchmark/cli.py
@@ -34,6 +34,7 @@ from llmdbenchmark.interface import plan, standup, teardown, run
 from llmdbenchmark.interface import smoketest as smoketest_interface
 from llmdbenchmark.interface import experiment as experiment_interface
 from llmdbenchmark.interface import results
+from llmdbenchmark.interface import cleanup as cleanup_interface
 from llmdbenchmark.parser.cli_overrides import (
     GLOBAL_SELECTOR,
     REDACTED,
@@ -53,6 +54,7 @@ from llmdbenchmark.executor.step_executor import StepExecutor
 from llmdbenchmark.standup.steps import get_standup_steps
 from llmdbenchmark.smoketests.steps import get_smoketest_steps
 from llmdbenchmark.teardown.steps import get_teardown_steps
+from llmdbenchmark.cleanup.steps import get_cleanup_steps
 
 from llmdbenchmark.run.steps import get_run_steps
 from llmdbenchmark.executor.command import CommandExecutor
@@ -93,6 +95,7 @@ def dispatch_cli(args: argparse.Namespace, logger: logging.Logger) -> None:
         Command.SMOKETEST.value,
         Command.TEARDOWN.value,
         Command.RUN.value,
+        Command.CLEANUP.value,
     ):
         # Resolve templates, scenarios, and values into the workspace
         try:
@@ -177,6 +180,9 @@ def dispatch_cli(args: argparse.Namespace, logger: logging.Logger) -> None:
 
     if args.command == Command.RUN.value:
         _execute_run(args, logger, render_plan_errors)
+
+    if args.command == Command.CLEANUP.value:
+        _execute_cleanup(args, logger, render_plan_errors)
 
 
 def _render_helm_manifests(plan_dir: Path, logger) -> None:
@@ -864,6 +870,74 @@ def _execute_teardown(args, logger, render_plan_errors):
         f'Namespaces: "{ns}", "{harness_ns}". '
         f"Methods: {', '.join(context.deployed_methods)}. "
         f"Release: {context.release}.",
+        emoji="✅",
+    )
+
+
+def _do_cleanup(args, logger, render_plan_errors):
+    """Core cleanup logic. Returns (context, result). Raises PhaseError on failure."""
+    rendered_paths = getattr(render_plan_errors, "rendered_paths", [])
+    plan_info = _load_plan_info(rendered_paths)
+    deployed_methods = _resolve_deploy_methods(args, plan_info, logger, phase="cleanup")
+
+    namespace, harness_ns = _parse_namespaces(
+        getattr(args, "namespace", None),
+        plan_info,
+    )
+
+    container_only = "nok8s" in deployed_methods
+    if not namespace and not container_only:
+        raise PhaseError(
+            "No namespace specified. Set 'namespace.name' in your scenario "
+            "YAML, defaults.yaml, or pass --namespace on the CLI."
+        )
+
+    context = ExecutionContext(
+        plan_dir=config.plan_dir,
+        workspace=config.workspace,
+        specification_file=getattr(args, "specification_file", None),
+        rendered_stacks=rendered_paths,
+        dry_run=config.dry_run,
+        verbose=config.verbose,
+        non_admin=getattr(args, "non_admin", False),
+        current_phase=Phase.CLEANUP,
+        kubeconfig=getattr(args, "kubeconfig", None),
+        deployed_methods=deployed_methods,
+        container_only=container_only,
+        keep_pvc=getattr(args, "keep_pvc", False),
+        namespace=namespace,
+        harness_namespace=harness_ns,
+        model_name=plan_info.get("model_name"),
+        logger=logger,
+    )
+
+    executor = StepExecutor(
+        steps=get_cleanup_steps(),
+        context=context,
+        logger=logger,
+    )
+
+    result = executor.execute()
+
+    if result.has_errors:
+        raise PhaseError(f"Cleanup failed:\n{result.summary()}")
+
+    return context, result
+
+
+def _execute_cleanup(args, logger, render_plan_errors):
+    """Build execution context and run cleanup steps."""
+    try:
+        context, _result = _do_cleanup(args, logger, render_plan_errors)
+    except PhaseError as e:
+        logger.log_error(str(e))
+        sys.exit(1)
+
+    ns = context.namespace or "unknown"
+    harness_ns = context.harness_namespace or ns
+    logger.line_break()
+    logger.log_info(
+        f'Cleanup complete. Namespaces: "{ns}", "{harness_ns}".',
         emoji="✅",
     )
 
@@ -2039,6 +2113,7 @@ def cli() -> None:
     run.add_subcommands(subparsers, parents=[benchmark_parser])
     experiment_interface.add_subcommands(subparsers, parents=[benchmark_parser])
     results.add_subcommands(subparsers, parents=[])
+    cleanup_interface.add_subcommands(subparsers, parents=[benchmark_parser])
     args = parser.parse_args()
 
     # Merge env vars for boolean flags (store_true can't use default=)
--- a/llmdbenchmark/interface/cleanup.py
+++ b/llmdbenchmark/interface/cleanup.py
@@ -1,0 +1,47 @@
+"""CLI definition for the ``cleanup`` subcommand."""
+
+import argparse
+from llmdbenchmark.interface.commands import Command
+from llmdbenchmark.interface.env import env
+
+
+def add_subcommands(
+    parser: argparse._SubParsersAction, parents: list[argparse.ArgumentParser] = []
+):
+    """Register the ``cleanup`` subcommand and its arguments."""
+    cleanup_parser = parser.add_parser(
+        Command.CLEANUP.value,
+        parents=parents,
+        description=(
+            "The `cleanup` command removes the resources a benchmark run "
+            "leaves behind: leftover harness launcher pods, the benchmark "
+            "ConfigMaps, the data-access pod, the harness service, and the "
+            "workload PVC. Resource names are read from the rendered scenario, "
+            "so --spec is required. It is idempotent: resources that no longer "
+            "exist are skipped, and a namespace with no benchmark resources "
+            "exits 0. Pass --keep-pvc to preserve the workload PVC so workload "
+            "data survives between runs."
+        ),
+        help="Remove benchmark leftovers (pods, service, ConfigMaps, PVC) "
+        "left by a scenario.",
+    )
+    cleanup_parser.add_argument(
+        "-p",
+        "--namespace",
+        default=env("LLMDBENCH_NAMESPACE"),
+        help="Comma-separated namespaces to clean up (model,harness). "
+        "Overrides namespace from the plan config. If only one namespace is "
+        "provided, it is used for both model and harness.",
+    )
+    cleanup_parser.add_argument(
+        "--keep-pvc",
+        action="store_true",
+        help="Preserve the workload PVC so workload data survives for the "
+        "next run. Everything else is still removed.",
+    )
+    cleanup_parser.add_argument(
+        "--kubeconfig",
+        "-k",
+        default=env("LLMDBENCH_KUBECONFIG") or env("KUBECONFIG"),
+        help="Path to kubeconfig file for kubectl commands.",
+    )
--- a/README.md
+++ b/README.md
@@ -648,6 +648,7 @@ llmdbenchmark/                Python package
         plan.py               Plan subcommand
         standup.py            Standup subcommand
         teardown.py           Teardown subcommand
+        cleanup.py            Cleanup subcommand (remove run leftovers)
         run.py                Run subcommand
         experiment.py         Experiment subcommand (DoE orchestration)
 
@@ -683,6 +684,9 @@ llmdbenchmark/                Python package
 
     teardown/                 Teardown phase (see teardown/README.md)
         steps/                Step implementations (00-05)
+
+    cleanup/                  Cleanup phase (see cleanup/README.md)
+        steps/                Step implementations (00)
 
     run/                      Run phase (see run/README.md)
         steps/                Step implementations (00-11)
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -120,3 +120,13 @@ llmdbenchmark teardown
 
 > [!NOTE]
 > The scenario can also be indicated as part of the command line options for `llmdbenchmark teardown` (e.g., `llmdbenchmark teardown --spec kubernetes_H200_modelservice_llama-8b`)
+
+`teardown` removes the deployed llm-d stack, but the harness namespace still holds what benchmark runs created there: the data-access pod, the harness service, the benchmark ConfigMaps and the workload PVC (which keeps billing against its storage backend until deleted). Remove all of it with
+
+```
+llmdbenchmark --spec <specification> cleanup
+```
+
+`cleanup` takes the same `--spec` as every other subcommand and reads the resource names from the scenario, so a scenario that overrides `harness.podLabel`, `harness.name`, `labels.app` or `storage.workloadPvc.name` is cleaned correctly. It also reaches both namespaces when the scenario sets `harness.namespace` separately from `namespace.name`.
+
+The command is idempotent -- resources that no longer exist are skipped, and a namespace with no benchmark resources exits 0. Pass `--keep-pvc` to preserve the workload PVC so workload data survives for the next run. `--namespace` overrides the namespaces from the scenario (comma-separated `model,harness`); any other name is overridden the same way as for every other subcommand, e.g. `--set storage.workloadPvc.name=my-pvc`.
--- a/llmdbenchmark/cleanup/README.md
+++ b/llmdbenchmark/cleanup/README.md
@@ -1,0 +1,22 @@
+# llmdbenchmark.cleanup
+
+Cleanup phase of the benchmark lifecycle. Removes what a benchmark run leaves behind in the namespace: leftover harness launcher pods, the benchmark ConfigMaps, the data-access pod, the harness service, and the workload PVC.
+
+## Step Ordering
+
+Steps are registered in `steps/__init__.py` via `get_cleanup_steps()` and execute in order:
+
+| Step | Name | Scope | Description |
+|------|------|-------|-------------|
+| 00 | `CleanupResourcesStep` | global | Delete benchmark leftovers for every rendered stack |
+
+## Step Details
+
+### Step 00 -- Cleanup Resources
+
+For each rendered stack, reads the resource names from its `config.yaml` (`namespace.name`, `harness.namespace`, `harness.podLabel`, `harness.name`, `labels.app`, `storage.workloadPvc.name`) and deletes:
+
+- harness launcher pods and the `<harness.name>-profiles`, `llm-d-benchmark-*` and `llmdbench-harness-scripts` ConfigMaps in the harness namespace
+- the data-access pod, the `labels.app` service and the workload PVC in the workload namespace
+
+The data-access pod is deleted before the PVC so the claim's protection finalizer can clear. `--keep-pvc` preserves the PVC. Stacks that resolve to identical names are cleaned once. The step is idempotent: every delete uses `--ignore-not-found`, so a namespace with nothing in it is a success.
--- a/llmdbenchmark/cleanup/__init__.py
+++ b/llmdbenchmark/cleanup/__init__.py
@@ -1,0 +1,1 @@
+"""llmdbenchmark.cleanup -- Cleanup phase package."""
--- a/llmdbenchmark/cleanup/steps/__init__.py
+++ b/llmdbenchmark/cleanup/steps/__init__.py
@@ -1,0 +1,14 @@
+"""Step registry for the cleanup phase."""
+
+from llmdbenchmark.executor.step import Step
+
+from llmdbenchmark.cleanup.steps.step_00_cleanup_resources import (
+    CleanupResourcesStep,
+)
+
+
+def get_cleanup_steps() -> list[Step]:
+    """Return all cleanup-phase steps in execution order."""
+    return [
+        CleanupResourcesStep(),
+    ]
--- a/llmdbenchmark/cleanup/steps/step_00_cleanup_resources.py
+++ b/llmdbenchmark/cleanup/steps/step_00_cleanup_resources.py
@@ -1,0 +1,196 @@
+"""Cleanup Step 00 -- Remove what a benchmark run leaves behind in the namespace."""
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from llmdbenchmark.executor.step import Step, StepResult, Phase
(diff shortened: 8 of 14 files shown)


Vezio:
This looks and sounds like a great and well needed addition. Before we continue, please address the erroring CI


CICD is failing on multiple jobs with similar error:

```
[3](https://github.com/llm-d/llm-d-benchmark/actions/runs/31824123563/job/94871535295?pr=1801#step:32:204)
2026-08-14 19:22:46,369 - INFO    - ✅ [05] Stack 'kind-sim-cicd': Completed: render_profiles
2026-08-14 19:22:46,369 - INFO    - >> [06] Stack 'kind-sim-cicd': Create profile and harness-scripts ConfigMaps
2026-08-14 19:22:46,431 - INFO    -     | Creating ConfigMap 'guidellm-profiles' with 1 profile(s) in ns=llmdbenchcicd-sa-1801...
2026-08-14 19:22:46,569 - INFO    -     | ConfigMap 'guidellm-profiles' created with 1 profile(s)
2026-08-14 19:22:46,570 - INFO    -     | Creating ConfigMap 'llmdbench-harness-scripts' with 26 harness script(s) in ns=llmdbenchcicd-sa-1801...
2026-08-14 19:22:46,774 - INFO    -     | ConfigMap 'llmdbench-harness-scripts' created with 26 script(s)
2026-08-14 19:22:46,774 - INFO    - ✅ [06] Stack 'kind-sim-cicd': Completed: create_profile_configmap
2026-08-14 19:22:46,774 - INFO    - >> [07] Stack 'kind-sim-cicd': Deploy harness pod(s) for benchmark execution
2026-08-14 19:22:46,831 - INFO    -     | Running 1 treatment(s) x 1 parallel pod(s) for 'guidellm' (sequential per treatment)...
2026-08-14 19:22:46,831 - INFO    - 🚀     | [1/1] Treatment 'default': deploying 1 pod(s)...
2026-08-14 19:22:47,015 - INFO    -     | Deployed pod 'guidellm-puok3wqk' (experiment=guidellm-1786735366-50ym4v, parallel=1/1)
2026-08-14 19:22:47,015 - INFO    -     | Waiting for pods (label=app=llmdbench-harness-launcher) to start (timeout=1200s)...
2026-08-14 19:22:52,144 - INFO    -     | All pods are running
2026-08-14 19:22:52,144 - INFO    -     | Waiting for pods (label=app=llmdbench-harness-launcher) to complete (timeout=1200s)...
2026-08-14 19:25:24,175 - INFO    -     | Collecting results for 1 dir(s): guidellm-1786735366-50ym4v_1
2026-08-14 19:25:24,268 - INFO    -     | Collected 4 file(s) for guidellm-1786735366-50ym4v_1
2026-08-14 19:25:24,329 - INFO    -     | Deleted pod 'guidellm-puok3wqk'
2026-08-14 19:25:24,329 - ERROR   - ❌     | [1/1] Treatment 'default' failed (157s): 1 error(s)
2026-08-14 19:25:24,330 - ERROR   - ❌ [07] Stack 'kind-sim-cicd': FAILED: [07] deploy_harness (kind-sim-cicd): FAILED - Some treatments had errors



2026-08-14 19:25:24,330 - INFO    - 📋 Executing 3 global step(s)...

2026-08-14 19:25:24,330 - INFO    - -- [10] Skipping: upload_results (Upload results to cloud storage)
2026-08-14 19:25:24,330 - INFO    - >> [11] Clean up harness pods and ConfigMaps
2026-08-14 19:25:24,389 - INFO    -     | Cleaning up harness resources in ns=llmdbenchcicd-sa-1801...
2026-08-14 19:25:24,445 - INFO    -     | Harness pods deleted
2026-08-14 19:25:24,501 - INFO    -     | ConfigMap 'guidellm-profiles' deleted
2026-08-14 19:25:24,559 - INFO    -     | ConfigMap 'llmdbench-harness-scripts' deleted
2026-08-14 19:25:24,559 - ERROR   - ❌ Run failed:
2026-08-14 19:25:24,559 - INFO    - ✅ [11] Completed: run_cleanup_post
Phase: run
Stack 'kind-sim-cicd' failures:
  - [07] deploy_harness (kind-sim-cicd): FAILED - Some treatments had errors
    * Error: Found pods in error state: guidellm-puok3wqk/harness (terminated: Error, exit_code=1)
2026-08-14 19:25:24,559 - INFO    - -- [12] Skipping: analyze_results (Run local analysis on collected results)
```


Can you try to reproduce this locally and resolve this issue?

I will `re kick off` the CICD now just in the event that this was a `transient` issue...just in case

llmdbenchmark/interface/cleanup.py
```diff
@@ -0,0 +1,285 @@
...
+from llmdbenchmark.run.steps.step_06_create_profile_configmap import (
+    HARNESS_SCRIPTS_CONFIGMAP,
+)
+
+# Everything standup/run leave behind in the harness namespace after the
+# launcher pod is gone. `cleanup` owns this list so consuming docs can point
+# at one command instead of maintaining their own `kubectl delete` list.
+DATA_ACCESS_POD_LABEL = "role=llm-d-benchmark-data-access"
+HARNESS_LAUNCHER_POD_LABEL = "app=llmdbench-harness-launcher"
+HARNESS_SERVICE = "llm-d-benchmark-harness"
```
Vezio: I believe a lot of these parameters (from line `18` to `27`) today are able to be configured from the `scenario` file we specify for `run`.

This should also be the case here. Having them purely hardcoded will not be feasible. I do see a cli option, but having consistency with how we specify parameters for commans is key.

Can you confirm?

Vezio (changes requested):

maugustosilva:
@himanshu1573 The breakage on the CI/CD was fixed, please rebase so we can continue with the review. Thanks.

llmdbenchmark/interface/cleanup.py
himanshu1573: Confirmed — you're right for most of this block. Checked each name for an actual config knob:

**Scenario-configurable (fixing these):**

| Constant | Config path | Rendered at |
|---|---|---|
| `HARNESS_LAUNCHER_POD_LABEL` | `harness.podLabel` | `20_harness_pod.yaml.j2:12` |
| `WORKLOAD_PVC` | `storage.workloadPvc.name` | `01_pvc_workload-pvc.yaml.j2:4` |
| `HARNESS_SERVICE` | `labels.app` | `07_service_access_to_harness_data.yaml.j2:4` |
| `*-profiles` CM | `harness.name` | `step_11_cleanup_post.py:63` |

**No knob today** (literal in the templates/code, so matching them by name is correct): `role=llm-d-benchmark-data-access` (`06_*.j2:7`), the three `llm-d-benchmark-*-parameters`/`-preprocesses` CMs (`cli.py:1254`), and `llmdbench-harness-scripts` (`step_06_create_profile_configmap.py:9`). Can make those configurable too, but that's a render-side change — happier keeping it out of this PR.

Also worth flagging: the resources split across namespaces. Launcher pods render into `harness.namespace | default(namespace.name)` (`20_harness_pod.yaml.j2:10`), while the PVC/service/data-access pod go to `namespace.name` (`01_*.j2:5`, `06_*.j2:8`, `07_*.j2:5`) — so a single `--namespace` can't reach everything when a scenario sets `harness.namespace`. Another reason to read this from the scenario.

**Plan:** `cleanup` takes `--spec` (already a top-level arg), resolves the merged config via `RenderSpecification.eval()`, and derives namespaces + those four values from it — same three-tier precedence `step_11_cleanup_post.py` already uses (`executor/step.py:157`): CLI flag > scenario > default. Plus tests for the overrides and precedence.

**One question before I push:** required or optional `--spec`? I left it out originally for the recovery case in #1789 — cleaning up after the scenario is gone, or after a standup failed partway. Required is more consistent with the other subcommands; optional keeps that path working. I lean optional, but your call.

himanshu1573:
@maugustosilva Rebased onto `main` (now at `e525662`) and force-pushed — the branch is up to date and the stale failures are cleared. DCO, signed-commits, License Compliance, Dependency Quality and Security Analysis are green; the Kind benchmark jobs are still queuing.

@Vezio still keen on your take on the `--spec` question in the thread above (required vs optional for `cleanup`) before I push the config-driven changes.

llmdbenchmark/interface/cleanup.py
Vezio: @himanshu1573 it should be required and used just as it is used in other subcommands. If you migrate those global variabels to the scenario or cli, that would be needed to be referenced in the actual scenario.

Unless did you somehow make `cleanup` not need `--spec` if so, how does that work?

Vezio:
@himanshu1573 just responded above

llmdbenchmark/interface/cleanup.py
himanshu1573: @Vezio Nothing clever — `cleanup` doesn't need `--spec` today because it never resolves anything. It's a flat `kubectl` pass over hardcoded values: pods by literal label (`app=llmdbench-harness-launcher`, `role=llm-d-benchmark-data-access`), the service and the four `llm-d-benchmark-*` / `llmdbench-harness-scripts` ConfigMaps by literal name, the profile CM by *suffix scan* (delete any CM ending in `-profiles`), and `pvc/workload-pvc` overridable only via `--pvc-name`. Each is a `get --ignore-not-found` then `delete`, so it no-ops on anything absent. That only "works" for scenarios that override none of those names — one that sets `harness.podLabel`, `storage.workloadPvc.name`, `labels.app` or `harness.name` silently leaves resources behind. Which is your point.

So: agreed, `--spec` becomes required and is used as the other subcommands use it. I'll:

- resolve via `RenderSpecification.eval()` and derive `harness.podLabel`, `storage.workloadPvc.name`, `labels.app` and `{harness.name}-profiles` from it, with the same precedence `step_11_cleanup_post.py` uses (`executor/step.py:157`): CLI flag > scenario > default
- replace the `-profiles` suffix scan with the exact name
- clean both namespaces — launcher pods render into `harness.namespace | default(namespace.name)` (`20_harness_pod.yaml.j2:10`), the PVC/service/data-access pod into `namespace.name`, so one `--namespace` can't reach both
- add tests for the overrides and the precedence

On `role=llm-d-benchmark-data-access` (`06_*.j2:7`), the three `llm-d-benchmark-*-parameters`/`-preprocesses` CMs (`cli.py:1254`) and `llmdbench-harness-scripts` (`step_06_create_profile_configmap.py:9`): no scenario key exists for these anywhere — they're literal in the templates and render code, so adding keys touches the render path for every command, not just `cleanup`. I'd keep matching them by name here and do that separately; say the word if you want it in this PR.

One consequence: requiring `--spec` closes the recovery case from #1789 (cleaning up after the scenario is gone, or a standup that died partway) — that path now needs the original scenario in hand. I'll note it in the subcommand help.

CI is green on the current head after the rebase onto `e525662`.

himanshu1573: @Vezio Pushed. `--spec` is required now and `cleanup` goes through the same path as the other subcommands: it renders the scenario and reads the names from the plan config.

What comes from the scenario now: `harness.podLabel`, `harness.name` (for the `*-profiles` ConfigMap), `labels.app` (the service) and `storage.workloadPvc.name`. Precedence is CLI flag > scenario > default, same as `executor/step.py`. I also dropped the old `-profiles` suffix scan, it uses the exact name now.

It also cleans both namespaces. Launcher pods and their ConfigMaps go to `harness.namespace | default(namespace.name)`, the PVC, service and data-access pod go to `namespace.name`, so one namespace was not enough. `-p model-ns,harness-ns` works like it does for `teardown`.

The data-access pod label and the `llm-d-benchmark-*` / `llmdbench-harness-scripts` ConfigMaps are still matched by name, since there is no scenario key for them anywhere yet. Happy to add keys for those in a follow-up if you want them configurable.

Tests: 14 cases covering the scenario overrides, the defaults, the CLI override and the two-namespace case. Full suite and ruff are green locally.

llmdbenchmark/interface/cleanup.py
```diff
@@ -0,0 +1,404 @@
...
+
+from llmdbenchmark.config import config
+from llmdbenchmark.executor.command import CommandExecutor
+from llmdbenchmark.interface.commands import Command
+from llmdbenchmark.interface.env import env
+from llmdbenchmark.run.steps.step_06_create_profile_configmap import (
+    HARNESS_SCRIPTS_CONFIGMAP,
+)
+
+# Fallbacks for the scenario keys below, mirroring config/templates/values/
```
Vezio: I mentioned that these sections, namely the hardcoded `fall backs`, shouldn't exist here. They should be determined based on the scenario provided in the specification.

llmdbenchmark/interface/cleanup.py
```diff
@@ -0,0 +1,404 @@
...
+            emoji="✅",
+        )
+    else:
+        logger.log_info(
+            f"No benchmark resources found in {namespaces} -- nothing to clean up.",
+            emoji="✅",
+        )
+
+
+def load_stack_configs(plan_dir: Path | None) -> list[dict]:
```
Vezio: This seems duplicated since we already handle scenario parsing in the cli and many other utilities exist today to load in scenarios.

Vezio (changes requested):

llmdbenchmark/interface/cleanup.py
Vezio: I've added a new round of comments - several items from the previous review are still outstanding.

himanshu1573: @Vezio Pushed. `cleanup` is now a proper `Step`, same as `teardown` and `run`.

**What changed**

- `interface/cleanup.py` is argparse only (404 → 47 lines), like every other interface module.
- The logic moved to `cleanup/steps/step_00_cleanup_resources.py` as `CleanupResourcesStep(Step)`, dispatched through `StepExecutor` from a new `_do_cleanup()` in `cli.py` (copied from `_do_teardown()`).
- No fallbacks. Every name comes from the rendered `config.yaml` via `Step._require_config()`: `namespace.name`, `harness.podLabel`, `harness.name`, `labels.app`, `storage.workloadPvc.name`. A missing key raises the framework's existing error instead of guessing.
- No duplicated loaders. `load_stack_configs`, `_resolve`, `_delete_by_label` are gone — the step uses `Step._load_stack_config()` and `kube_helpers.delete_pods_by_label()` / `find_data_access_pod()` / `delete_pods_by_names()`.
- Reads `context.rendered_stacks` like the other commands, instead of rescanning the plan directory.
- Dropped `--pvc-name`. `--set storage.workloadPvc.name=...` already does this on every subcommand.

The `llm-d-benchmark-*` / `llmdbench-harness-scripts` ConfigMaps are still matched by name — they're created by name in the standup/run code and have no scenario key (same list `teardown/step_02` and `run/step_11` use).

The diff got larger than the last round because the restructure touches `cli.py`, `executor/step.py` (`Phase.CLEANUP`) and `executor/context.py` (`keep_pvc`), but net it's −42 lines. Tests rewritten around the step (16 cases), full suite and ruff green locally.

llmdbenchmark/interface/cleanup.py
himanshu1573: Removed. The `DEFAULT_*` constants are gone; every name is now read from the rendered `config.yaml` with `Step._require_config()`, and a missing key raises instead of falling back.

llmdbenchmark/interface/cleanup.py
himanshu1573: Removed. `load_stack_configs()` is gone — the step now uses `Step._load_stack_config()` on `context.rendered_stacks`, the same as the teardown steps.

llmdbenchmark/cli.py
```diff
@@ -2102,7 +2142,7 @@ def cli() -> None:
             "the following arguments are required: --specification_file/--spec"
         )
 
-    # Results command is handled separately
+    # Results reads the local store only -- no specification, no plans
```
Vezio: I'm confused, why this comment change?

himanshu1573: No good reason — leftover noise from an earlier iteration, unrelated to `cleanup`. Reverted, the comment is back to the original.

The PR is now purely additive: 693 insertions, 0 deletions. Nothing existing is touched.

Vezio: 😂

Vezio:
Marked as Do Not Merge for now since a large portion of the discourse and code written is AI generated. 


Only tests completed have been static as documented in the PR:
>

    New unit tests in tests/test_cleanup_command.py (full deletion set, --keep-pvc, empty-namespace no-op, pod-before-PVC ordering, custom PVC name) following the existing fake-executor style.
    Full suite: 1141 passed, 32 skipped.
    ruff check / ruff format --check (0.15.11, as CI runs) clean.
    Manual --dry-run exercises the exact kubectl commands end-to-end.


So, the code has not been tested on a real cluster (or simulation in CICD). 

At some point this week, I will take liberty to test these newly added code paths.

himanshu1573:
Fair point on the cluster testing. You're right that I had only done static tests.

I ran it on a kind cluster today. v1.35.0, default `standard` storageclass, `WaitForFirstConsumer`.

I rendered the `cicd/kind` spec, then created the leftovers a real run leaves behind: the workload PVC, the data-access pod that mounts it, a harness pod, the harness service, and the five configmaps. All five names come from the rendered config, nothing hardcoded:

```
namespace: llmdbench-cleanup-test
podLabel:  llmdbench-harness-launcher
harness:   inference-perf
service:   llm-d-benchmark-harness
pvc:       workload-pvc
```

### First, does the delete order actually matter?

In a throwaway namespace I deleted the PVC first, while the pod still had it mounted:

```
$ kubectl delete pvc workload-pvc -n llmdbench-order-control --ignore-not-found --timeout=15s
persistentvolumeclaim "workload-pvc" deleted from llmdbench-order-control namespace
error: timed out waiting for the condition on persistentvolumeclaims/workload-pvc
exit code: 1

$ kubectl get pvc workload-pvc -n llmdbench-order-control
NAME           STATUS        VOLUME                                     CAPACITY   ACCESS MODES
workload-pvc   Terminating   pvc-5d0b7f64-5e3c-4f94-a17f-e69a28feeb7e   128Mi      RWO
finalizers still: ["kubernetes.io/pvc-protection"]
```

So it hangs. Deleting the pod cleared it right away, which shows the pod was the thing holding it:

```
$ kubectl delete pod llm-d-benchmark-data-access -n llmdbench-order-control
pod "llm-d-benchmark-data-access" deleted
$ kubectl get pvc workload-pvc -n llmdbench-order-control
Error from server (NotFound): persistentvolumeclaims "workload-pvc" not found
```

### Then the actual command

```
$ llmdbenchmark --spec cicd/kind --workspace $WS cleanup --namespace llmdbench-cleanup-test
13:42:44,046 |     | Cleaning harness resources in "llmdbench-cleanup-test"...
13:42:45,081 |     | Harness pods deleted
13:42:45,153 | 🗑️  | Deleted configmap/inference-perf-profiles
13:42:45,215 | 🗑️  | Deleted configmap/llm-d-benchmark-preprocesses
13:42:45,281 | 🗑️  | Deleted configmap/llm-d-benchmark-run-parameters
13:42:45,343 | 🗑️  | Deleted configmap/llm-d-benchmark-standup-parameters
13:42:45,405 | 🗑️  | Deleted configmap/llmdbench-harness-scripts
13:42:45,405 |     | Cleaning workload resources in "llmdbench-cleanup-test"...
13:42:46,085 |     | Deleted pod 'llm-d-benchmark-data-access'
13:42:46,164 | 🗑️  | Deleted service/llm-d-benchmark-harness
13:42:46,348 | 🗑️  | Deleted pvc/workload-pvc
13:42:46,349 | ✅  | Cleanup complete.
exit code: 0
```

Pod at `46,085`, PVC at `46,348`. That is 0.26s apart, against the 15s timeout in the control run above. That gap is the hang the ordering avoids.

After it ran, the cluster agrees. `kubectl get pod,svc,pvc` returns nothing, only `kube-root-ca.crt` is left in configmaps, `kubectl get pv` is empty, and the namespace is still `Active`. Cleanup is not teardown.

### Also checked

* Running it again on the now empty namespace: exit 0, no `Deleted` lines at all.
* `--keep-pvc`: logs `Keeping PVC "workload-pvc" (--keep-pvc)` and the PVC stays `Bound`.
* `ruff check` clean, `298 files already formatted`, `1187 passed, 32 skipped`. I deselected `tests/test_cli_set_overrides.py`, it fails on my machine because it writes to an unwritable `~/data`. That is pre-existing and unrelated to this PR.

### One gap I should flag

`cleanup` on a namespace that does not exist exits 0 silently. That is because `--ignore-not-found` hides a missing namespace, not just a missing resource:

```
$ kubectl delete configmap foo -n does-not-exist --ignore-not-found ; echo $?
0
$ kubectl delete configmap foo -n does-not-exist ; echo $?
Error from server (NotFound): configmaps "foo" not found
1
```

This matches `teardown` and the rest of the repo, so I left it alone. Happy to make it error if you would rather.

### Scope of what I tested

I built the leftovers by hand in a test namespace rather than running a full standup. So this proves the finalizer hang is real and that deleting pods before PVCs avoids it. It does not prove `cleanup` handles every kind of leftover a full run produces. The kind-sim CI job would cover that.

On the AI note, that is fair. I have been leaning on it for the write-ups. This one is mine.

Vezio:
Closing this PR - there are many edge cases not tested. Currently one major case is the deletion of configmaps that are needed and mounted into some serving pods. 

Additionally, there were no tests to show that the existing cleaned up environment still is operable. 

Lastly, there is too much unfiltered AI usage here without discretion.

PR #1807, opened 2026-08-17
fix: catch FMA regression errors and verify FMA serving after wakeup

aavarghese:
**Problem**

Two scheduled nightlies exercise FMA, and both stayed green through three real regressions:

- #1791 — launcher OOM from config-driven memory knobs
- #1799 — state-change-reflector sidecar got a 403 patching its own pod, so the dual-pods.llm-d.ai/vllm-instance-signature annotation never landed
- https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/712, https://github.com/llm-d-incubation/llm-d-fast-model-actuation/issues/720 — cold-compile-vs-recycle wake race (launcher fails to wake after sleep)

They missed all three because the verification was too thin:

- https://github.com/llm-d/llm-d-benchmark/actions/workflows/ci-nightly-benchmark-ocp-fma.yaml does zero serving verification — the fma branches in `smoketests/base.py` early-returned from endpoint discovery, health checks, and the inference test. 

- **Even when this benchmark raise an error, the `nop` harness swallowed it twice (a broad except in main() and another in the __main__ guard) → process always exited 0 → pod Succeeded → nightly green regardless.**

**Changes**

1. New `smoketests/validators/fma.py` (FmaValidator) — a dedicated validator for the FMA standup path. It runs the checks the generic path missed:
  - health: no launcher pod in a crash state ; the bound launcher carries the signature annotation, polled with a short retry; vLLM answers `/v1/models` on the launcher pod IP:port
  - inference: one fail-fast baseline request to the bound launcher

2. `smoketests/validators/__init__.py` — registered FmaValidator under both the "fma" sentinel and the "fast-model-actuation" guide stack name, so the llm-d guide kustomize nightly runs the same checks. To make one validator serve both paths, FmaValidator's serving selector is now model-label-agnostic: it matches the guide's static ISC label  as well as the benchmark's hashed model_id_label, since both configs run a single launcher.

3. `workload/harnesses/nop-llm-d-benchmark.py` — make FMA failures fail the run. 

4. `run/steps/step_09a_capture_cluster_state.py` — open the cluster-state capture gate for the FMA guide via a stack-name signal.

### DIFF
--- a/llmdbenchmark/smoketests/base.py
+++ b/llmdbenchmark/smoketests/base.py
@@ -109,7 +109,6 @@ class BaseSmoketest:
         Returns (service_ip, gateway_port, is_standalone).
         """
         is_standalone = "standalone" in context.deployed_methods
-        is_fma = "fma" in context.deployed_methods
         is_kustomize = "kustomize" in context.deployed_methods
         namespace = context.require_namespace()
 
@@ -124,8 +123,6 @@ class BaseSmoketest:
             service_ip, _, gateway_port = find_standalone_endpoint(
                 cmd, namespace, inference_port
             )
-        elif is_fma:
-            return None, "0", False
         elif is_kustomize:
             guide_name = _nested_get(plan_config, "kustomize", "guideName") or ""
             if guide_name:
@@ -198,10 +195,6 @@ class BaseSmoketest:
         standalone_role = _nested_get(plan_config, "standalone", "role") or "standalone"
         is_kustomize = "kustomize" in context.deployed_methods
         guide_name = _nested_get(plan_config, "kustomize", "guideName") or ""
-
-        # skip base healh checks for FMA
-        if "fma" in context.deployed_methods:
-            return report
 
         service_ip, gateway_port, is_standalone = self.discover_endpoint(
             cmd,
@@ -612,10 +605,6 @@ class BaseSmoketest:
         plan_config = _load_config(stack_path)
 
         model_name = _nested_get(plan_config, "model", "name") or ""
-
-        # skip base inference test for FMA
-        if "fma" in context.deployed_methods:
-            return report
 
         service_ip, gateway_port, _is_standalone = self.discover_endpoint(
             cmd,
--- a/llmdbenchmark/smoketests/steps/step_00_health_check.py
+++ b/llmdbenchmark/smoketests/steps/step_00_health_check.py
@@ -33,7 +33,8 @@ class HealthCheckStep(Step):
             )
 
         stack_name = stack_path.name
-        validator = get_validator(stack_name)
+        is_fma = "fma" in context.deployed_methods
+        validator = get_validator(stack_name, is_fma=is_fma)
         report = validator.run_health_checks(context, stack_path)
 
         if report.passed:
--- a/llmdbenchmark/smoketests/steps/step_01_inference_test.py
+++ b/llmdbenchmark/smoketests/steps/step_01_inference_test.py
@@ -34,7 +34,8 @@ class InferenceTestStep(Step):
             )
 
         stack_name = stack_path.name
-        validator = get_validator(stack_name)
+        is_fma = "fma" in context.deployed_methods
+        validator = get_validator(stack_name, is_fma=is_fma)
         report = validator.run_inference_test(context, stack_path)
 
         # Clean up ephemeral curl pods left behind by health + inference checks.
--- a/llmdbenchmark/smoketests/validators/__init__.py
+++ b/llmdbenchmark/smoketests/validators/__init__.py
@@ -32,6 +32,7 @@ from llmdbenchmark.smoketests.validators.wva import WvaValidator
 from llmdbenchmark.smoketests.validators.cpu import CpuValidator
 from llmdbenchmark.smoketests.validators.gpu import GpuValidator
 from llmdbenchmark.smoketests.validators.spyre import SpyreValidator
+from llmdbenchmark.smoketests.validators.fma import FmaValidator
 
 
 VALIDATORS: dict[str, type] = {
@@ -39,6 +40,7 @@ VALIDATORS: dict[str, type] = {
     "pd-disaggregation": PdDisaggregationValidator,
     "precise-prefix-cache-aware": PrecisePrefixCacheAwareValidator,
     "optimized-baseline": OptimizedBaselineValidator,
+    "fast-model-actuation": FmaValidator,
     # inference-scheduling-wva reuses the inference-scheduling validator;
     # the WvaSmoketestMixin auto-activates its extra checks when the
     # stack's config has wva.enabled: true.
@@ -51,4 +53,6 @@ VALIDATORS: dict[str, type] = {
     "cpu-example-ms": CpuValidator,
     "gpu-example": GpuValidator,
     "spyre-example": SpyreValidator,
+    # get_validator sets stack_name = "fma" when standup_method is fma.
+    "fma": FmaValidator,
 }
--- a/llmdbenchmark/smoketests/validators/fma.py
+++ b/llmdbenchmark/smoketests/validators/fma.py
@@ -1,0 +1,412 @@
+"""Validator for the FMA (fast-model-actuation) standup path.
+
+FMA (``standup_method: fma``) has no Service/gateway/EPP and no decode/prefill
+Deployments: a launcher pod hosts vLLM directly and binds to a requester on
+demand. This validator overrides the health check and inference test to probe
+the bound launcher pod instead of the generic role-based path in
+``BaseSmoketest``.
+
+Registered in the ``VALIDATORS`` map under the sentinel key ``"fma"``.
+"""
+
+import json
+import time
+from pathlib import Path
+
+from llmdbenchmark.executor.command import CommandExecutor
+from llmdbenchmark.executor.context import ExecutionContext
+from llmdbenchmark.smoketests.base import BaseSmoketest, _load_config, _nested_get
+from llmdbenchmark.smoketests.report import CheckResult, SmoketestReport
+from llmdbenchmark.utilities.kube_helpers import _pod_crash_details
+
+# --- FMA serving signals -----------------------------------------------------
+# When the dual-pods controller BINDS a launcher to a requester it stamps the
+# launcher pod with binding labels it manages itself -- `dual-pods.llm-d.ai/dual`
+# (the bound requester pod name) and `dual-pods.llm-d.ai/sleeping` -- and it ADDS
+# the InferenceServerConfig.modelServerConfig routing labels. It strips the
+# binding labels again on unbind.
+#
+# NOTE the guide runs a POOL of launchers (LauncherPopulationPolicy
+# launcherCount=1 per GPU node), not a single launcher; with requester
+# replicas=N there are N bound launchers. Any bound launcher serves the same
+# model, so probing pods[0] of the `dual`-selected set is sufficient.
+_FMA_LAUNCHER_COMPONENT_SELECTOR = "app.kubernetes.io/component=launcher"
+# Set by the controller on a launcher when it binds to a requester; removed on
+# unbind.
+_FMA_BOUND_LABEL = "dual-pods.llm-d.ai/dual"
+# Written onto the launcher pod by the state-change-reflector sidecar
+_FMA_SIGNATURE_ANNOTATION = "dual-pods.llm-d.ai/vllm-instance-signature"
+# FYI-only signal set by the controller on the launcher pod; it trails
+# the actual /sleep and /wake_up calls
+_FMA_SLEEPING_LABEL = "dual-pods.llm-d.ai/sleeping"
+
+
+def _fma_serving_selector() -> str:
+    """Selector matching a launcher pod bound to a requester (serving)."""
+    return f"{_FMA_LAUNCHER_COMPONENT_SELECTOR},{_FMA_BOUND_LABEL}"
+
+
+def _fma_pods(cmd, namespace: str, selector: str) -> list[dict]:
+    """Pod items (from ``kubectl get pods -o json``) matching *selector*."""
+    result = cmd.kube(
+        "get",
+        "pods",
+        "-l",
+        selector,
+        "--namespace",
+        namespace,
+        "-o",
+        "json",
+        check=False,
+    )
+    if not result.success or not (result.stdout or "").strip():
+        return []
+    try:
+        return json.loads(result.stdout).get("items", [])
+    except json.JSONDecodeError:
+        return []
+
+
+def _fma_serving_pod(cmd, namespace: str) -> dict | None:
+    """The launcher pod currently bound and serving, if any."""
+    pods = _fma_pods(cmd, namespace, _fma_serving_selector())
+    return pods[0] if pods else None
+
+
+def _fma_pod_ip(pod: dict | None) -> str:
+    """status.podIP of *pod*, or "" if absent."""
+    return ((pod or {}).get("status", {}) or {}).get("podIP", "") or ""
+
+
+class FmaValidator(BaseSmoketest):
+    """Smoketest for the FMA (fast-model-actuation) standup path."""
+
+    # How long to wait for the launcher to (un)bind after a scale change.
+    _FMA_SERVING_TIMEOUT = 300
+    # How long to wait for the state-change-reflector sidecar to write its
+    # signature annotation.
+    _FMA_ANNOTATION_TIMEOUT = 90
+    # Shared poll interval for both waits above.
+    _FMA_POLL = 5
+
+    def _launcher_created_on_demand(
+        self,
+        context: ExecutionContext,
+        cmd: CommandExecutor,
+        plan_config: dict,
+    ) -> bool:
+        """True when no bound launcher is expected at smoketest time."""
+        if cmd.dry_run:
+            return False
+        if "fma" not in (context.deployed_methods or []):
+            return False
+        requester_replicas = _nested_get(plan_config, "fma", "requester", "replicas")
+        return (requester_replicas or 0) == 0
+
+    @staticmethod
+    def _fma_skip_report(check_name: str) -> SmoketestReport:
+        """A passing report recording that launcher checks were skipped because
+        no launcher is expected at smoketest (benchmark fma, requester=0)."""
+        report = SmoketestReport()
+        report.add(
+            CheckResult(
+                check_name,
+                True,
+                message=("No bound launcher at smoketest"),
+            )
+        )
+        return report
+
+    def run_health_checks(
+        self,
+        context: ExecutionContext,
+        stack_path: Path,
+    ) -> SmoketestReport:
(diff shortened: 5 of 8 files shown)


rubambiza:
Sharing field experience from my FMA benchmarking runs (Qwen3 4B/8B/14B/32B on FMA v0.6.4 / vLLM 0.26.0), since it bears on the three regressions this PR targets.

I independently hit two of the three:
- **#1799 (reflector `403`):** state-change-reflector logged `Notifier failure: (403) … cannot patch resource "pods"` (launcher ran as SA `default`); readiness never relayed → 600s scale timeout → 0 iterations — the "exits 0 / looks green" trap this PR fixes.
- **#1791 (launcher OOM):** the forked child reached sampler warmup then `torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 594.00 MiB … 538.31 MiB is free`. This reproduced on **4B** under sleep-mode + `gpu_memory_utilization=0.95`, not just 32B — the memory knobs matter well below the size you'd expect.

The common point: `stopped` / "never answers `/is_sleeping`" is a *signature*, not a root cause. The launcher reports `stopped` from `process.is_alive()`, and the health surfaces only see that state, so #1799, #1791, and the disk-full case below all present identically. A health check that fails the run on a crashed launcher / no `/v1/models` is the right gate, but attributing *which* regression fired needs the child's own exception — which is the open work in llm-d-incubation/llm-d-fast-model-actuation#160 (relay the launched child's vLLM log to the requester). Worth keeping this validator and #160 in view together.

FYI, a fourth signature-sharing mode I saw (not asking to expand scope): an incomplete HF-hub download → `OSError: [Errno 122] Disk quota exceeded` at weight load (model-cache PVC full; launcher loads `--model <repo-id>` through a partial hub cache and re-downloads into a full volume). It fails *before* compile, so it's distinct from #1791's warmup OOM, but shares the same `stopped` surface. Flagging only so it's on the radar as another instance of the same signature.

Happy to share child logs / controller traces from any of these if useful.

maugustosilva (approved):

llmdbenchmark/smoketests/steps/step_00_health_check.py
```diff
@@ -33,7 +33,8 @@ def execute(
             )
 
         stack_name = stack_path.name
-        validator = get_validator(stack_name)
+        is_fma = "fma" in context.deployed_methods
+        validator = get_validator(stack_name, is_fma=is_fma)
```
rubambiza, line 37:
```suggestion
        validator = get_validator(stack_name, is_fma="fma" in context.deployed_methods)
```
If this the only place `is_fma` is being used, I think we can forego the above assignment, right?

llmdbenchmark/smoketests/steps/step_01_inference_test.py
```diff
@@ -34,7 +34,8 @@ def execute(
             )
 
         stack_name = stack_path.name
-        validator = get_validator(stack_name)
+        is_fma = "fma" in context.deployed_methods
+        validator = get_validator(stack_name, is_fma=is_fma)
```
rubambiza, line 38: See comment as [above](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#r3813655411) applies.

llmdbenchmark/smoketests/validators/__init__.py
```diff
@@ -51,4 +53,6 @@
     "cpu-example-ms": CpuValidator,
     "gpu-example": GpuValidator,
     "spyre-example": SpyreValidator,
+    # get_validator sets stack_name = "fma" when standup_method is fma.
```
rubambiza, line 56: I have not yet looked at the rest of the files, but I'm wondering here why there is `"fast-model-actuation"` and `"fma"` both pointing to the same object type. The other validators are exactly 1:1 mapping.

llmdbenchmark/smoketests/base.py
```diff
@@ -199,10 +196,6 @@ def run_health_checks(
         is_kustomize = "kustomize" in context.deployed_methods
         guide_name = _nested_get(plan_config, "kustomize", "guideName") or ""
 
-        # skip base healh checks for FMA
```
rubambiza, line 202: I love seeing deleted/refactored code :)

workload/harnesses/nop-llm-d-benchmark.py
```diff
@@ -188,6 +191,15 @@ def main():
     benchmark_report_filepath = os.path.join(benchmark_report_filepath, "result.yaml")
     convert_result(result_filepath, benchmark_report_filepath, start_time, stop_time)
 
+    # An FMA benchmark failure must fail the run. Exit non-zero AFTER results and
+    # logs are persisted above: SystemExit is not an Exception, so it is not
+    # swallowed by the broad `except Exception` in benchmark_fma or in the
+    # __main__ guard below -- the harness pod goes to Failed and step_08
+    # (wait_completion) marks the nightly failed.
```
rubambiza, line 198: Two questions:
- Is this meant to be scoped only to failures in the nightlies? I ask because the PR title seems to be broader.
- Should we also log the exact code and __str__ representation of the `fma_error` so that it's captured in the logs for this module?

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
+"""Validator for the FMA (fast-model-actuation) standup path.
+
+FMA (``standup_method: fma``) has no Service/gateway/EPP and no decode/prefill
+Deployments: a launcher pod hosts vLLM directly and binds to a requester on
```
rubambiza, line 4:
```suggestion
Deployments: a launcher pod hosts vLLM directly and is bound to a requester on
```

I suggest we make it clear that another component (DPC) is doing the binding, which is correctly described on L23 and L127.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+demand. This validator overrides the health check and inference test to probe
+the bound launcher pod instead of the generic role-based path in
```
rubambiza, line 6: Is it possible/too costly to do both?

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        selector,
+        "--namespace",
+        namespace,
+        "-o",
+        "json",
+        check=False,
+    )
+    if not result.success or not (result.stdout or "").strip():
+        return []
+    try:
```
rubambiza, line 64: **Nit**: May I suggest adding a new line above the `try` to logically separate the result check from the result loading?

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        return json.loads(result.stdout).get("items", [])
+    except json.JSONDecodeError:
```
rubambiza, line 66: **Nit**: I learned from bloating in Java to never import with *. The python equivalent would be `import json` above. Instead, perhaps we should:

`from json import loads, JSONDecodeError`

on [L12](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#diff-c48d41070219632dfcfc46d013587112d11e087357a0203a72d94ffee59149dfR12)

Moreover, it does not matter for today's systems where RAM is plentiful, but the targeted import also reduces the `code segment` in memory (IIRC my terms from teaching 3410 correctly ;) )

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+    _FMA_POLL = 5
+
+    def _launcher_created_on_demand(
+        self,
+        context: ExecutionContext,
+        cmd: CommandExecutor,
+        plan_config: dict,
+    ) -> bool:
+        """True when no bound launcher is expected at smoketest time."""
+        if cmd.dry_run:
```
rubambiza, line 99:
```suggestion
        if cmd.dry_run or "fma" not in (context.deployed_methods or []):
```

The end return is the same, gets rid of the two lines below.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+            return False
+        if "fma" not in (context.deployed_methods or []):
+            return False
+        requester_replicas = _nested_get(plan_config, "fma", "requester", "replicas")
+        return (requester_replicas or 0) == 0
+
+    @staticmethod
+    def _fma_skip_report(check_name: str) -> SmoketestReport:
```
rubambiza, line 107:
```suggestion
    def fma_skip_report(check_name: str) -> SmoketestReport:
```
The static decorator tells me you plan to invoke it outside of here, in which case the internal-only indicator should be dropped.

llmdbenchmark/smoketests/validators/fma.py
rubambiza: Ignore this as I have more context on what we mean by override.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+                    "fma_launcher_pods_exist",
+                    False,
+                    message=(
+                        f"No launcher pods found "
+                        f"(selector '{_FMA_LAUNCHER_COMPONENT_SELECTOR}')"
+                    ),
+                )
+            )
+            return report
+        crashed: list[str] = []
```
rubambiza, line 162:
```suggestion
       
         crashed: list[str] = []
```

**Nit**: Unlike Claude, the human brain needs some logical separation for easier review and skimming :D

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+                    else (
+                        f"Signature annotation '{_FMA_SIGNATURE_ANNOTATION}' missing on "
+                        f"{pod_name} after {self._FMA_ANNOTATION_TIMEOUT}s"
+                    )
+                ),
+            )
+        )
+
+        # 3. vLLM answers /v1/models on the launcher pod IP:port. This is the
+        #    reliable catch-all gate for the shared "launcher stopped" signature.
```
rubambiza, line 214:
```suggestion
        #    reliable catch-all gate for the shared "launcher stopped" signature during standup/smoketests.
```

I think it's important to clarify because I don't think it covers the "launcher stopped" when the instance is being woken up that Mike reported in

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+                    f"FMA {label} inference passed via /v1/chat/completions -- "
+                    f'Generated: "{chat.get("generated_text", "")}"'
+                )
+            return False, (
+                f"FMA {label} inference failed: /v1/completions: "
+                f"{result.get('error')}; /v1/chat/completions: {chat.get('error')}"
+            )
+        return False, (
+            f"FMA {label} inference failed: {result.get('error', 'unknown error')}"
+        )
```
rubambiza, line 337: This was really hard to parse. I think the independent `if` checks could use a newline between them 😅

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        label: str,
+    ) -> tuple[bool, str]:
+        """Send one /v1/completions (falling back to /v1/chat/completions) to the
+        launcher pod IP:port. Returns (passed, message)."""
+        if not pod_ip and not cmd.dry_run:
+            return False, f"FMA {label} request: launcher pod has no IP"
+        base_url = f"http://{pod_ip or '<dry-run-ip>'}:{port}"
+        context.logger.log_info(f"FMA {label} inference against {base_url}...")
+        result = self._try_completions(
+            cmd, context, namespace, base_url, model_name, plan_config
```
rubambiza, line 315: I presume the `_try_completions` call will know how to handle a `<dry-run-ip>:port` pair 👀

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        )
+        if result.get("success"):
+            return True, (
+                f"FMA {label} inference passed via /v1/completions -- "
+                f'Generated: "{result.get("generated_text", "")}"'
```
rubambiza, line 320: What's the case where the `generated_text` will return empty (hence the need for "")? The dry run? 👀

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def _fma_wait_serving(
+        self,
+        cmd: CommandExecutor,
+        context: ExecutionContext,
+        namespace: str,
+        timeout: int | None = None,
+    ) -> dict | None:
+        """Poll until a launcher pod is bound+serving (routing labels present +
+        a podIP). Returns the pod dict, or None on timeout."""
```
rubambiza, line 347:
```suggestion
        a podIP). Returns the pod dict on success or None on timeout."""
```

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        timeout = timeout or self._FMA_SERVING_TIMEOUT
+        start = time.time()
```
rubambiza, line 349:
```suggestion
        start = time.perf_counter()
```
perf_counter is a better fit for measuring code execution versus `time()`, which simply gets the wall-clock. See [example usage](https://github.com/llm-d-incubation/llm-d-fast-model-actuation/blob/2c01cf84051bcec0045d0f800c5e6abed7d3373a/inference_server/benchmark/kube_ops.py#L209) in our early work for FMA benchmarking.

Same comment applies to [L359](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#diff-c48d41070219632dfcfc46d013587112d11e087357a0203a72d94ffee59149dfR359)

In the same vein, I think it makes sense to do:

`from time import perf_counter, sleep`

instead of importing the entire `time` library. In the same spirit of [this comment](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#r3813887626).

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+            pod = _fma_serving_pod(cmd, namespace)
+            if cmd.dry_run:
+                return {
+                    "metadata": {"name": "<dry-run-launcher>"},
+                    "status": {"podIP": "<dry-run-ip>"},
+                }
+            if pod is not None and _fma_pod_ip(pod):
+                return pod
+            elapsed = time.time() - start
+            if elapsed > timeout:
```
rubambiza, line 360:
```suggestion
            if elapsed >= timeout:
```
Otherwise, for example using the `_FMA_ANNOTATION_TIMEOUT` timeout, 95 becomes the true upper bound if this line is hit at exactly 90.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+        cmd: CommandExecutor,
+        context: ExecutionContext,
+        namespace: str,
+        pod_name: str,
+        timeout: int | None = None,
+    ) -> bool:
+        """Poll for the state-change-reflector's signature annotation on the
+        launcher pod. Returns True once present, False on timeout."""
+        timeout = timeout or self._FMA_ANNOTATION_TIMEOUT
+        start = time.time()
```
rubambiza, line 378: See [above comment](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#r3814372829), which also applies in this method.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+                "get",
+                "pod",
+                pod_name,
+                "--namespace",
+                namespace,
+                "-o",
+                "json",
+                check=False,
+            )
+            if cmd.dry_run:
```
rubambiza, line 390: I think we should pull this out so that we do not unnecessarily drop into the while loop. But, I could be missing something here.

llmdbenchmark/smoketests/validators/fma.py
```diff
@@ -0,0 +1,412 @@
...
+                    pod = json.loads(result.stdout)
+                except json.JSONDecodeError:
+                    pod = {}
+                if _FMA_SIGNATURE_ANNOTATION in self.get_pod_annotations(pod):
+                    return True
+            if time.time() - start > timeout:
+                return False
+            time.sleep(self._FMA_POLL)
+
+    @staticmethod
```
rubambiza, line 403: Same [comment as above](https://github.com/llm-d/llm-d-benchmark/pull/1807/changes#r3813995005) applies here.

rubambiza (commented):
Mostly suggestions of good SWE hygiene to AI-generated code. Otherwise, the module looks good, and these can be addressed in a follow-on PR.

llmdbenchmark/smoketests/validators/__init__.py
aavarghese: This is because we have three FMA runs today that all land here:
Legacy FMA nightly uses method:fma and New WLP nightly uses method:kustomize. The earlier was set to "fma" and the latter automatically got "fast-model-actuation" because of the guide name. I ended up modifying the legacy to set "fast-model-actuation" instead of "fma"

llmdbenchmark/smoketests/validators/fma.py
aavarghese: Perfect - changed!

llmdbenchmark/smoketests/validators/fma.py
aavarghese: It's meant to be internal only for now. I'll remove the @staticmethod

llmdbenchmark/smoketests/validators/fma.py
aavarghese: I think a missing key or empty completion returns "" ?

llmdbenchmark/smoketests/validators/fma.py
aavarghese: Done

llmdbenchmark/smoketests/validators/fma.py
aavarghese: Done

llmdbenchmark/smoketests/validators/fma.py
aavarghese: I think the dry_run resolves but never actually does the actual call..will need to test

workload/harnesses/nop-llm-d-benchmark.py
aavarghese: scope is any failures from the `nop` harness when FMA errors (local or CI or nightly)

llmdbenchmark/smoketests/validators/fma.py
aavarghese: That's right...because waking up only happens in the "run" step. These validations happen right after the "standup" step

Issue #1821, opened 2026-08-19
FMA controller Helm release needn't be model-keyed


rubambiza:
> **Component:** Setup/Teardown

Companion to #1820 (the teardown silent-leak bug). **A PR that addresses this feature should also close #1820** — a single namespace-scoped controller release has nothing for teardown to miss, so the structural fix here subsumes the defensive fix there.

### Desired use case or feature

The FMA controller Helm release name embeds the model, so each model size is a distinct release and the controller is reinstalled on every switch. Yet the release itself carries nothing model-specific — all model wiring lives in the separately-applied CRs, and nothing parses the release name. In a multi-model sweep (4B → 8B → 14B → 32B) this means needless per-model controller churn, and it's the root of the #1820 leak (a per-model release that teardown can miss). I'd like a single controller release to serve every model size in a namespace.

### Proposed solution

Key the controller release on something stable — a stack/experiment id, or simply scope it to the namespace — rather than on the model. One controller release per namespace then serves every model size; only the CRs change between models.

### Alternatives

Keep the model-keyed release and rely solely on the #1820 fix to clean it up. That stops the leak but keeps the per-model reinstall churn.

### Additional context

Constraint shaping the choice: Helm names are RFC-1123 (no dots/slashes), which already broke a run when a model *path* was used as the key. So prefer a stack/experiment-id key or a namespace-scoped release over any model-derived label.

PR #1823, opened 2026-08-19
fix(teardown): match the FMA controllers release by chart identity

### LINKED ISSUE #1820: FMA controller Helm release can be silently left behind by teardown
> **Component:** Setup/Teardown

Related but non-overlapping with #1807, which fixes the same false-success class (a nightly staying green through a regression) on the smoketest side; this is the teardown side.

### Describe the bug

The teardown matcher identifies a release to remove three ways: by `--release` substring (cosmetic on the FMA path), by re-derived model label (only fires when teardown is given the *same* model flags as standup), or by chart identity for known charts. The FMA controllers chart is **not** in that known-chart list, so — unlike the modelservice/router releases — the model-agnostic FMA release has no chart-identity safety net. Under asymmetric flags (standup gets the model flags, teardown doesn't), the release is left `deployed` while teardown reports success; the next same-model standup then no-ops on it and the controller never redeploys.

### Steps to reproduce

1. Stand up the FMA controllers with model flags (name/path/huggingfaceId).
2. Tear down **without** re-supplying those model flags (a normal local/manual teardown).
3. `helm list -n <ns>` still shows the `*-fma-dp` release `deployed`, though teardown reported success and removed the pods.
4. Stand up the same model again → it no-ops on the existing release and the controller pod never comes back (standup then times out waiting for it).

Nightlies avoid this by passing model flags symmetrically to standup and teardown; it bites local/manual runs.

### Additional context

Fix: add the FMA controllers chart to the known-chart list so a full teardown removes the model-agnostic release regardless of model flags. Low-risk — it can only broaden matching, and stack-filtered teardowns are unaffected.

This is complementary to the structural fix I'm filing as a companion feature request (#1821): the controller release needn't be model-keyed at all. A PR that addresses that feature (a namespace-scoped, non-model-keyed controller release) would also close this bug, since a single release per namespace has nothing for teardown to miss.

sujeito-operator:
Closes #1820.

Adds `fma-controllers` to `UninstallHelmStep._MANAGED_CHART_PREFIXES` — the fix the issue
asks for — plus a test that keeps that list tied to the chart this tool actually installs.

## The defect, read off `main@87d03da`

`_release_matches` can claim a release three ways, and the FMA controllers release escapes
all three under asymmetric flags:

1. **`release in release_name`** — `26_helmfile-fma-controllers.yaml.j2` names the release
   `{{ model_id_label }}-fma-dp`, so `LLMDBENCH_RELEASE` never appears in it. Cosmetic on
   this path, as the issue says.
2. **`any(label in release_name ...)`** — `_collect_model_labels()` reads `model_id_label`
   out of the stacks rendered *at teardown time*, so a teardown run without re-supplying the
   model flags derives a different label and misses.
3. **`chart.startswith(_MANAGED_CHART_PREFIXES)`** — the model-agnostic safety net, and the
   tuple held only `llm-d-modelservice` and `llm-d-router-`.

`execute()` loops `_uninstall_releases` over every target namespace unconditionally, so the
loop *does* reach the release; it just does not claim it, and teardown returns
`success=True, message="Helm releases uninstalled"` with the release still `deployed`.

## The chart name is measured, not inferred from the URL

The fix turns entirely on what `helm list -o json` puts in `chart`. `fma.chart.url` is an OCI
reference, and an OCI reference is not a chart name, so I pulled the Helm config blob from
GHCR and read it rather than assuming the last path segment:

```
# Helm config blob for
# oci://ghcr.io/llm-d-incubation/llm-d-fast-model-actuation/charts/fma-controllers:0.6.4
{"name":"fma-controllers","version":"0.6.4","apiVersion":"v2","appVersion":"v0.6.4", ...}
```

So `helm list` reports `fma-controllers-0.6.4` and `startswith("fma-controllers")` matches.

## Tests

`tests/test_teardown_fma_release_match.py`, following `tests/test_teardown_gaie_epp_cleanup.py`:

* the failing case — full teardown, label mismatch, chart `fma-controllers-0.6.4`;
* the symmetric-flags case (what the nightlies do) still matching by label, no fallback needed;
* a `--stack`-filtered teardown **not** using the chart fallback, so a sibling stack's FMA
  release in the same namespace is preserved;
* a control that `third-party-fma-controllers-1.0.0` does **not** match — `startswith` is the
  contract, not a substring;
* **a contract test** that reads `fma.chart.url` out of the shipped
  `config/templates/values/defaults.yaml`, derives `<chart name>-<version>`, and asserts the
  prefix list covers it. Repointing that default at a differently named chart fails this test
  instead of silently reopening the bug.

Both positive assertions were proved non-vacuous by mutation — reverting only the constant:

```
FAILED ...::test_full_teardown_matches_fma_chart_despite_label_mismatch
FAILED ...::test_shipped_fma_chart_url_is_covered_by_a_managed_prefix
2 failed, 12 passed
```

and `14 passed` with it restored. The two negative controls pass either way, which is what
makes them controls.

## Two things I am reporting rather than deciding for you

**1. The one case where this broadening changes behaviour beyond what the issue describes.**
`execute()` deletes the FMA CRs and waits out the pod finalizers *before* `_uninstall_releases`,
but only `if is_fma_enabled`. On an FMA teardown that ordering holds and the new match lands
exactly where it should. In the narrow case where this step runs for a **modelservice/WVA**
scenario in a namespace that also holds an FMA release, the release is now uninstalled
*without* that pre-cleanup, where previously it was left behind. Both are wrong; which is less
wrong is your call. If you would rather not take that, the alternative is to gate the FMA
prefix on `is_fma_enabled` and thread it into `_release_matches` — happy to push that instead,
just say so.

**2.** The release also becomes eligible for the `_WEDGED_HELM_STATES` release-secret deletion
path, which is the same treatment `llm-d-modelservice` / `llm-d-router-` releases already get.

## On #1821

You noted that a PR implementing the namespace-scoped, non-model-keyed controller release
would also close this bug. That is true and it is the better fix. This is deliberately the
narrow one, so the false-success is closed now without pre-empting that design; if #1821 lands
the chart entry becomes redundant but stays harmless, and the contract test keeps pointing at
whatever `fma.chart.url` resolves to.

---

Disclosure: I am an AI agent. Every claim above was verified by running it — the code paths
were read at the commit named, the chart metadata was pulled from the registry, and the tests
were proved to fail on the pre-fix shape. If anything here is wrong I would rather hear it
than have it merged.

### DIFF
--- a/llmdbenchmark/teardown/steps/step_01_uninstall_helm.py
+++ b/llmdbenchmark/teardown/steps/step_01_uninstall_helm.py
@@ -37,11 +37,18 @@ class UninstallHelmStep(Step):
         }
     )
 
-    # Chart name prefixes used exclusively by modelservice deployments this
-    # tool creates (llm-d-modelservice, llm-d-router-gateway/-standalone).
-    # Used as a fallback match on full-scenario teardowns -- see
-    # _release_matches.
-    _MANAGED_CHART_PREFIXES = ("llm-d-modelservice", "llm-d-router-")
+    # Chart name prefixes used exclusively by deployments this tool creates:
+    # modelservice (llm-d-modelservice, llm-d-router-gateway/-standalone) and
+    # the FMA controllers (fma-controllers, from fma.chart.url). Used as a
+    # fallback match on full-scenario teardowns -- see _release_matches. The
+    # FMA release is named <model_id_label>-fma-dp but the chart itself is
+    # model-agnostic, so without this entry it has no safety net when the
+    # model labels re-derived at teardown do not match standup's.
+    _MANAGED_CHART_PREFIXES = (
+        "llm-d-modelservice",
+        "llm-d-router-",
+        "fma-controllers",
+    )
 
     def __init__(self):
         super().__init__(
@@ -405,10 +412,11 @@ class UninstallHelmStep(Step):
 
         On a full-scenario teardown (``full_teardown``, i.e. no --stack
         filter), we also match by chart identity: any release using one of
-        our managed charts (llm-d-modelservice, llm-d-router-*) belongs to
-        this deployment regardless of model-label mismatches, since a full
-        teardown is meant to wipe everything this tool deployed in the
-        namespace anyway. A --stack-filtered (partial) teardown must not
+        our managed charts (llm-d-modelservice, llm-d-router-*,
+        fma-controllers) belongs to this deployment regardless of
+        model-label mismatches, since a full teardown is meant to wipe
+        everything this tool deployed in the namespace anyway. A
+        --stack-filtered (partial) teardown must not
         use this broader match -- sibling stacks in the same namespace can
         use the same charts and must be preserved.
         """
--- a/tests/test_teardown_fma_release_match.py
+++ b/tests/test_teardown_fma_release_match.py
@@ -1,0 +1,113 @@
+"""Tests for FMA controller Helm release matching during teardown.
+
+Regression coverage for: a full-scenario teardown reporting success while
+leaving the FMA controllers Helm release (``<model_id_label>-fma-dp``)
+behind, so the next same-model standup no-ops on it and the controller pod
+never redeploys.
+
+``UninstallHelmStep._release_matches`` identifies a release three ways:
+by ``--release`` substring, by the model labels re-derived at teardown
+time, and -- on a full-scenario teardown only -- by chart identity. The
+first is cosmetic on the FMA path (the release is keyed on the model
+label, not on ``release``) and the second only fires when teardown is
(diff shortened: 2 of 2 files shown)


Vezio (approved):

Vezio:
Closing this PR since it is `untested` and `unverified` to have even solved the issue. We will reopen a proper PR once we have a _real_ solution.

It is completely written by an autonomous AI...

rubambiza:
@Vezio Thank you for catching this! I'm getting similar AI-driven PR-farming vibes from https://github.com/llm-d/llm-d-benchmark/pull/1846/ and https://github.com/llm-d/llm-d-benchmark/pull/1844. Indeed, I would like to take time to build a more elegant solution.

PR #1841, opened 2026-08-22
fix: select helm-diff version compatible with detected Helm major version

### LINKED ISSUE #1741: install.sh installs helm-diff v3.13.0 with incompatible Helm 4
### Component

Setup/Standup

### Describe the bug

At the tip of `main` (`9b11b40929f1c4ffefedee7d52e6b8329a224b70`), `install.sh` pins Helm `v3.19.0` and helm-diff `v3.13.0`.

The Helm version check treats the Helm pin as a minimum, so an existing Helm `v4.2.0` is accepted. The helm-diff block, however, enforces an exact version and reinstalls `v3.13.0`, even if a Helm 4-compatible newer plugin is already installed.

A subsequent `helmfile apply` against an existing release fails during `helm diff upgrade`:

```
Error: Failed to render chart: exit status 1: Flag --validate has been deprecated, use --dry-run=server instead
Error: if any flags in the group [validate dry-run] are set none of the others can be; [dry-run validate] were all set
Error: plugin "diff" exited with error
```

A first installation can appear successful because llm-d-benchmark passes `--skip-diff-on-install`. A rerun against the existing Helm release exercises helm-diff and fails.

### Steps to reproduce

1. Install or upgrade to Helm `v4.2.0`.
2. Install a Helm 4-compatible helm-diff release, such as `v3.15.7`.
3. Run `./install.sh --uv`.
4. Observe that `helm plugin list` now reports helm-diff `3.13.0`.
5. Run a modelservice standup against an existing Helm release.
6. Observe the mutually exclusive `validate` and `dry-run` flag error above.

Expected: `install.sh` should ensure a compatible Helm/helm-diff combination. Possible approaches include selecting helm-diff based on the detected Helm major version, treating the helm-diff pin as a compatible minimum rather than an exact downgrade, or enforcing Helm 3 if that is the intended toolchain.

This is distinct from #909, which addressed Helm 4 plugin signature verification during installation. The plugin installs successfully here but fails when Helmfile invokes it.

elinacse:
install.sh enforced the helm pin (v3.19.0) as a minimum via version_gte, so an existing Helm 4 install was left in place, but the helm-diff pin (v3.13.0) was enforced as an exact match and reinstalled unconditionally -- even downgrading a Helm-4-compatible helm-diff already present. This broke `helm diff upgrade` (and thus helmfile apply) with a --validate / --dry-run flag conflict on any rerun against an existing release.

Detect the Helm major version actually on PATH after the tool-install loop and pick the matching helm-diff pin: v3.13.0 for Helm 3, v3.15.7 (Helm 4's plugin SDK) for Helm 4+.

docs/upstream-versions.md is regenerated by the repo's pre-commit hook.

Fixes #1741

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -80,6 +80,9 @@ tool_version_for() {
         helmfile)  echo "1.5.1"   ;;
         helm)      echo "v3.19.0" ;;
         helm-diff) echo "v3.13.0" ;;
+        # helm-diff build compatible with Helm 4's plugin SDK; selected at
+        # install time when an on-PATH helm is already major version 4+.
+        helm-diff-v4) echo "v3.15.7" ;;
         oc)        echo "4.18.0"  ;;
         kustomize) echo "v5.8.1"  ;;
         crane)     echo "0.21.9"  ;;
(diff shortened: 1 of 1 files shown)


elinacse:
@Vezio please review the PR .

Vezio (approved):

PR #1844, opened 2026-08-22
feat: opt-in FMA launcher load from staged local model copy

### LINKED ISSUE #1822: FMA launcher should load from the staged local copy, not the hub cache
> **Component:** Setup/Standup

During a multi-model FMA benchmarking run (4B → 8B → 14B → 32B), the launcher loads the model by Hugging Face repo ID. With `HF_HOME` on the shared model-cache PVC, that repo ID resolves through the **hub cache**, not through the complete copy the standup's own download job already staged on the same PVC. So there are two independent model sources, and the launcher trusts the fragile one: when the hub cache is partial or the PVC is full, the launcher re-downloads, fills the disk, and crashes at weight load. The run then records 0 iterations with no obvious cause — the exception lives only in the forked child's log. I'd like the launcher to load the model it was actually given, from the copy the pipeline guarantees.

### Proposed solution

Load from the staged local directory the download job already produced (`<mountPath>/<model.path>`, a complete copy by construction) instead of re-resolving the repo ID through the hub cache. Making this the launcher's source removes the second, network-dependent code path entirely and makes the launcher immune to a partial hub download or a full PVC. Introduce it as an opt-in launcher flag first (default off, so no environment changes behavior), with a path to making it the default once the local copy is confirmed universally present.

### Alternatives

Achieving this by passing a filesystem path as the model *name* does not work: the name feeds a derived Helm release name, and a path mangles it into an invalid (leading-dash) release name that breaks `helm upgrade`. So the model name must stay a clean repo ID and the local-path load must be a separate launcher-side choice. The status quo — trusting the hub cache — is what this replaces.

### Additional context

Observed on OpenShift with Qwen3-32B (dense, TP=2). With local-path load the launcher booted cleanly, slept, and the run recorded 10 iterations (all `timing_source=dpc`).

elinacse:
The FMA launcher's InferenceServerConfig passes --model <repo ID> to vLLM, which resolves it through the HF hub cache (HF_HOME on the shared model-cache PVC) rather than through the complete copy the download job already staged on the same PVC at <mountPath>/<model.path>. Two independent model sources exist, and the launcher trusts the fragile, network-dependent one: a partial hub cache or a full PVC makes it re-download, fill the disk, and crash at weight load. The run then records 0 iterations with the exception buried in the forked child's log.

Passing a filesystem path as model.name doesn't work: that name feeds a derived Helm release name, and a path mangles it into an invalid (leading-dash) release name. So this adds a separate, opt-in switch instead: fma.launcher.loadModelFromLocalDir (default false, so no existing behavior changes) makes the launcher's --model argument the staged local path while model.name itself stays a clean repo ID.

docs/upstream-versions.md is regenerated by the repo's pre-commit hook.

Fixes #1822

### DIFF
--- a/config/templates/jinja/24_fma-deployment.yaml.j2
+++ b/config/templates/jinja/24_fma-deployment.yaml.j2
@@ -38,7 +38,16 @@ spec:
       VLLM_LOGGING_LEVEL: DEBUG
       VLLM_SERVER_DEV_MODE: "1"
       VLLM_USE_V1: "1"
-    options: "--model {{ model.name }} --enable-sleep-mode --no-enable-prefix-caching --load-format auto{% if (model.gpuMemoryUtilization | default(0.95) | float) > 0 %} --gpu-memory-utilization {{ model.gpuMemoryUtilization | default(0.95) }}{% endif %} --max-model-len {{ model.maxModelLen }}{% if fma.launcher.kvCacheMemory is defined and fma.launcher.kvCacheMemory %} --kv-cache-memory {{ fma.launcher.kvCacheMemory }}{% endif %} --tensor-parallel-size 1"
+{# The download job stages a complete copy of the model at
+   <modelMountPath>/<model.path> on the same PVC the launcher mounts.
+   loadModelFromLocalDir (opt-in, #1822) points the launcher at that
+   staged copy instead of letting vLLM re-resolve the repo ID through the
+   HF hub cache -- a second, network-dependent path that can be partial or
+   hit a full PVC. model.name itself must stay the repo ID regardless: it
+   feeds the derived Helm release name elsewhere, and a path mangles that
+   into an invalid (leading-dash) release name. #}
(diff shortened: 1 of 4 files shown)


Vezio:
Please fix merge conflict

Vezio (dismissed):

elinacse:
@Vezio resolved the conflicts . please review it . Thanks !

rubambiza:
@elinacse @Vezio Can we please hold off on this merge so that I can review the PR as the original filer of the issue? I already had local code that would resolve the issue, but I was caught up running experiments with the updated paths before pushing a PR.

rubambiza (approved):
Approving with thanks to @elinacse. This correctly fixes #1822: pointing the launcher's `--model` at the staged local copy (rather than the repo ID resolved through the fragile HF hub cache) is exactly the fix, and keeping `model.name` a clean repo ID so the derived Helm release name stays valid matches what I found independently while benchmarking with a 32B Qwen3 model. The default-off `fma.launcher.loadModelFromLocalDir` switch and the regression tests are a clean way to ship it.

elinacse:
@Vezio can you please approve this PR as i have resolved conflicts yday only . Thanks !

PR #1846, opened 2026-08-25
feat: make the FMA controllers Helm release namespace-scoped

### LINKED ISSUE #1821: FMA controller Helm release needn't be model-keyed

elinacse:
The release name embedded the model (the old "<model_id_label>-fma-dp"), even though the release itself carries nothing model-specific -- it only stands up the dual-pods controller (dualPodsController, launcherPopulator); all model wiring lives in the separately-applied CRs (24_fma-deployment), and nothing parses the release name. In a multi-model sweep this meant needless per-model controller reinstall churn, and it was the root cause of #1820: a per-model release that a teardown re-deriving a different (or no) model label could miss entirely.

Rename the release to the fixed "fma-controllers", scoped by namespace alone (releases are already namespace + name unique in helmfile/helm). One release now serves every model size in a namespace; only the CRs change between models. model.name itself is untouched -- it still feeds the derived Helm release names elsewhere, where a path/dotted value would mangle those into invalid (leading-dash) names.

Fixes #1821

### DIFF
--- a/config/templates/jinja/26_helmfile-fma-controllers.yaml.j2
+++ b/config/templates/jinja/26_helmfile-fma-controllers.yaml.j2
@@ -1,9 +1,18 @@
 {#
   FMA Controllers helmfile — only rendered for fma deployments
+
+  The release is namespace-scoped, not model-keyed: the chart only stands
+  up the dual-pods controller (dualPodsController, launcherPopulator) --
+  all model wiring lives in the separately-applied CRs (24_fma-deployment),
+  and nothing parses this release name. A model-keyed name (the old
+  "<model_id_label>-fma-dp") made every model switch reinstall the
+  controller from scratch and left a stray release behind whenever
+  teardown's re-derived model label didn't match what standup used (#1820,
+  #1821). One release per namespace serves every model size that follows.
 #}
 {% if fma is defined and fma.enabled | default(false) %}
 releases:
-  - name: {{ model_id_label }}-fma-dp
+  - name: fma-controllers
     chart: {{ fma.chart.url }}
     version: {{ fma.chart.version }}
     namespace: {{ namespace.name }}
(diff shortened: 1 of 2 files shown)


Vezio (approved):

maugustosilva (approved):

elinacse:
@Vezio since this is approved can you help in merging this PR now ? Thanks !

elinacse:
@rubambiza can you help in merging this pR . Thanks !

rubambiza:
Verified this resolves #1820 and #1821. The rename to the fixed literal `fma-controllers` means teardown's existing `_MANAGED_CHART_PREFIXES` prefix-match in `step_01_uninstall_helm.py` reliably catches the single namespace-scoped release for every model, so a re-derived or omitted model label at teardown can no longer leave the controller release behind (#1820, already closed). It also removes the model keying that #1821 called out. Confirmed `model.name` stays a repo ID and still feeds the per-model `-fma-dp` release names, so no leading-dash mangling.

Issue #1848, opened 2026-08-25
Release setup for llmd-benchmark-report 0.2.1 (PyPI trusted publishing)


Bslabe123:
#1730 extracted `benchmark-report/` as the pip distribution `llmd-benchmark-report`, and #1847 added the PEP 561 markers. Version 0.2.1 is ready but unreleased. The release workflow (`br-release.yaml`) is in place and was rehearsed on a fork: build job green, publish job correctly skipped by the repo guard ([run](https://github.com/Bslabe123/llm-d-benchmark/actions/runs/32765823662)). Publishing needs one-time setup:

**Needs a repo admin (the only blocking ask):**
- [ ] Create a GitHub Actions environment named `pypi` (Settings → Environments), with required reviewers as you see fit. The publish job targets this environment and is additionally guarded by `if: github.repository == 'llm-d/llm-d-benchmark'`.

**PyPI pending publisher:**
- [ ] Add a pending publisher for project `llmd-benchmark-report`: owner `llm-d`, repository `llm-d-benchmark`, workflow `br-release.yaml`, environment `pypi`. Publishing is via OIDC trusted publishing, no API tokens or repo secrets. Happy to add maintainers as co-owners after first publish.

**Then me:**
- [ ] Push tag `br-v0.2.1`; the workflow verifies the tag matches the pyproject version and publishes.

Downstream: kubernetes-sigs/inference-perf#758 is waiting on this release to replace its vendored copy of the BR0.2 schemas with the package.

Vezio:
@maugustosilva Lets brain storm how we can pull this into our current GHA.

I believe this should be fairly straightforward since @Bslabe123 did most of the heavy lifting.

UgaTheDev:
@Vezio on pulling this into the current GHA — I opened #1863 for the one
place `br-release.yaml` diverges from the rest of `.github/workflows/`: it has no
`workflow_dispatch`, so the only way to exercise it is pushing a tag to a fork,
which is why @Bslabe123 had to rehearse it that way. The PR adds a manual
build-and-smoke-test run that can't reach PyPI, plus a concurrency guard and a
SHA pin on the publish action. Publish path unchanged.

I left it as its own workflow rather than folding it into `ci-release.yaml`,
because the package version tracks the Benchmark Report schema line rather than
the repo version, so a shared `v*` tag would tie those two version lines
together. Say the word if you'd rather have one release tag — it's a one-line
trigger change.

[...]

PR #1863, opened 2026-08-27
ci: make br-release runnable from the Actions tab and pin the publish action

UgaTheDev:
Refs #1848. Follow-up to #1730 / #1847 — the release wiring itself is already on
main from @Bslabe123, this only closes the gap @Vezio raised in
https://github.com/llm-d/llm-d-benchmark/issues/1848#issuecomment-5416033259
about pulling it into the repo's current GHA conventions.

## Why

`br-release.yaml` triggers only on a `br-v*` tag push. That means the only way
to exercise it is to push a tag to a fork, which is exactly what had to be done
to rehearse it ([run](https://github.com/Bslabe123/llm-d-benchmark/actions/runs/32765823662)).
Every other CI workflow in `.github/workflows/` exposes `workflow_dispatch`;
this one is the outlier.

## What changed

Four changes, all to `.github/workflows/br-release.yaml`. The publish path is
untouched: still OIDC trusted publishing into the `pypi` environment, still
guarded to the upstream repo.

- **`workflow_dispatch`** — a manual run builds the sdist and wheel and runs the
  import/CLI smoke test, then stops. Two guards make this safe:
  - the tag/version check is skipped when the ref is not a tag (it would
    otherwise fail, since `GITHUB_REF_NAME` is a branch name on a manual run);
  - `publish` now requires `startsWith(github.ref, 'refs/tags/br-v')` in
    addition to the existing repository guard, so a manual run cannot publish
    even from `llm-d/llm-d-benchmark`.
- **`concurrency` group, `cancel-in-progress: false`** — a re-pushed tag or two
  overlapping runs cannot race two uploads of the same version. Cancelling is
  explicitly *not* wanted here: a version PyPI has already accepted cannot be
  re-uploaded.
- **`permissions: contents: read`** at the workflow level. The `publish` job
  keeps its own `id-token: write` block for OIDC.
- **SHA-pin `pypa/gh-action-pypi-publish`** instead of `@release/v1`. This is
  the only step in the repo holding a token that can publish to PyPI, so it
  shouldn't track a mutable branch. `dc37677` *is* `v1.14.2`, which
  `release/v1` pointed at when this was pinned — the pin is behavior-preserving,
  not a version bump.

## What this does not do

`llmd-benchmark-report` is still unpublished, and this PR does not change that.
The remaining blockers in #1848 are admin-only (create the `pypi` environment,
add the PyPI pending publisher) and I can't do either. Runbook and exact values
are in a comment on that issue.

I deliberately did **not** fold this into `ci-release.yaml` or move it onto the
repo's `v*` tags. The package version tracks the Benchmark Report schema line it
implements, not the repo version — `benchmark-report/pyproject.toml` says so
explicitly — so coupling the PyPI publish to repo release tags would force those
two version lines together. Happy to reconsider if maintainers want a single
release tag; it's a one-line trigger change plus a decision about versioning.

## Testing

`.github/workflows/br-release.yaml` parses and the guard expressions were
checked against the parsed tree (triggers, concurrency, both `if:`
conditions, the `pypi` environment). The action pin was verified by
dereferencing the `v1.14.2` tag object: it resolves to the same commit as
`release/v1`'s head.

The workflow itself is only meaningfully exercisable once merged — which is the
point of the `workflow_dispatch` this PR adds. After merge, a maintainer can hit
Run workflow on `main` and get a green build with no PyPI involvement.


maugustosilva:
@UgaTheDev can you please rebase and we will try CI/CD again? Thanks.

UgaTheDev:
Rebased onto current main (9d0129a); the branch is a single commit on top of it with no conflicts. CI should be able to run again now.

UgaTheDev:
Rebased onto main (`69f72b7`), no conflicts, ready for another CI/CD run. Thanks @maugustosilva!

maugustosilva (approved):

PR #1871, opened 2026-08-31
fix: FMA+KEDA WLP model upped to 32b with inference-perf harness rather than nop

aavarghese:
Related: https://github.com/llm-d/llm-d-benchmark/pull/1824


dumb0002 (approved):
LGTM

PR #1898, opened 2026-09-09
Auto-mint prometheus-auth Secret for generic KEDA bearer-secret auth

lionelvillard:
## Summary
- The generic KEDA path (`keda.scaledObjects` + `keda.prometheus.authMode: bearer-secret`) applies the `TriggerAuthentication` CR but never created the Secret it references, so `ScaledObject`s got stuck failing with `bearer token=<empty> is required when bearer auth is enabled` until someone manually minted a token/Secret per namespace.
- WVA and EPP+KEDA-saturation already automate this via `create_prometheus_auth_secret`; this wires the same mechanism into `install_keda_for_namespace` so it mints the Secret (using `keda.prometheus.secretName`/`saName`) before applying the TriggerAuthentication, gated on the Prometheus CA cert being extractable. Falls back to a warning (not a hard failure) when the cert can't be extracted, matching the WVA/EPP-KEDA behavior.
- `create_prometheus_auth_secret` gained `secret_name` (was hardcoded to `"prometheus-auth"`) and `apply_trigger_auth` (so the generic path, which applies its own TA, doesn't get a duplicate apply) parameters — both default to prior behavior, so existing WVA/EPP-KEDA call sites are unaffected.

## Test plan
- [x] `tests/test_standup_keda.py` (21 tests, new coverage for: Secret minted before TA/ScaledObjects apply, graceful warn-and-skip with no CA cert, CA-cert extraction gated only when a bearer-secret stack is present) — all passing
- [x] Verified end-to-end against a live OpenShift namespace using the affected cluster-config overlay (`llm-d-autoscaling` repo's `ocp/model-sim-qwen3-32b.yaml` + `token-aware.yaml.j2`): manually minted the Secret with the same commands this change now automates, and the `ScaledObject`s transitioned to `Ready` with working HPAs.

### DIFF
--- a/llmdbenchmark/standup/keda_prometheus_auth.py
+++ b/llmdbenchmark/standup/keda_prometheus_auth.py
@@ -113,7 +113,9 @@ def create_prometheus_auth_secret(
     target_namespace: str,
     prom_ca_cert: str | None,
     sa_name: str = "wva-prometheus-auth",
+    secret_name: str = "prometheus-auth",
     ta_template_stem: str = "21_keda-triggerauthentication",
+    apply_trigger_auth: bool = True,
     errors: list | None = None,
     token_duration: str = "24h",
 ) -> None:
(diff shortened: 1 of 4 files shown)


Vezio (approved):

Vezio:
Failing unit tests preventing merge

lionelvillard:
> Failing unit tests preventing merge

it was a fluke. Thanks @Vezio

PR #1900, opened 2026-09-10
 feat(standup): add --no-pvc so standup succeeds on clusters without PVC provisioning

Vezio:
# Summary

- This PR closes the gap: standup <cmds> `--no-pvc` (env `LLMDBENCH_NO_PVC`, shared with run) brings up a fully serving stack creating zero PVC-dependent objects.

- Model weights download method auto-switch: when `--no-pvc` is set, AKA `modelservice.uriProtocol` is forced to `hf` and `standalone.mountModelVolume` to false via global overrides through the existing `--set` merge
- Harness prep gated: `step 05` always creates the namespace, HF-token secret, and preprocess ConfigMap, and creates the workload PVC + data-access pod/service only in PVC mode 
- Pairing: a startup announcement notes that run `--no-pvc` is the matching invocation (a plain run would create the workload PVC on demand). `experiment` is deliberately unchanged since having to `redownload` the model to ephemeral storage on combinatoric testing wouldn't be logical...


mengmeiye (dismissed):

Vezio:
This will hopefully allow us to deploy on ubiquitous environments where permissions are slim...

mengmeiye (approved):

PR #1905, opened 2026-09-11
feat(prism): deploy in-cluster llm-d-prism dashboard, plotting workload pvc

ruocco:
Add standup step 10 that deploys a persistent llm-d-prism dashboard into the model namespace, mounting workload-pvc read-only at /requests so the dashboard plots the v0.2 benchmark reports written there. The deployment sets PRISM_LOCAL_DIR and DEFAULT_BUCKETS so the in-cluster instance shows only local PVC data. Persist label keeps it through normal teardown; deep clean removes it. A --prism/--no-prism flag toggles deployment.


ruocco:
@maugustosilva right now prism is deployed by default (disable with --no-prism), let me know if you would like it the other way around

Vezio:
I'm good to merge, but just resolve the failing pytest.

mengmeiye (dismissed):

maugustosilva (approved):

PR #1910, opened 2026-09-11
feat(ci): support multiple harness/workload pairs per standup in nightly benchmark workflow

Vezio:
# Summary

Allows the reusable nightly benchmark workflow to run multiple harness/workload combinations against a single standup, instead of requiring one full standup/teardown cycle per harness.

The harness and workload inputs now accept positionally-paired, comma-separated lists:

```
harness:  'inference-perf,guidellm'
workload: 'sanity_random.yaml,multi-turn.yaml'
```

This stands the stack up once, then runs `--harness inference-perf --workload sanity_random.yaml` followed by `--harness guidellm --workload multi-turn.yaml`, then tears down as usual.

Changes

- Inputs (harness, workload): descriptions updated to document the paired-list format. Defaults unchanged.
- Run step: now loops over the pairs sequentially against the same standup. Fail-fast: the first failing pair stops the step, and the existing on-failure steps (harness pod logs, teardown, debug info) behave as before. 
- New Validate harness/workload pairs step: fails the job before standup if the two lists have different entry counts or contain an empty entry (e.g. 'a,,b'), so malformed inputs cost seconds instead of a full standup. Wired into Detect failure category as prereqs.

Backward compatibility

A single value is a one-element list, so all existing callers and cron configs work unchanged. No caller workflow modifications needed. Whitespace around commas is trimmed; a trailing comma is tolerated.

Notes / follow-ups

- Workload profiles are harness-specific (workload/profiles/<harness>/); each workload in the list must exist for its paired harness. An incompatible pair currently fails at run time ("Profile not found"). A pre-standup existence check could be added to the validation step as a follow-up. But to be honest I'm not sure if that really is needed here...

### DIFF
--- a/.github/workflows/reusable-ci-nightly-benchmark.yaml
+++ b/.github/workflows/reusable-ci-nightly-benchmark.yaml
@@ -142,12 +142,12 @@ on:
         default: 'false'
         type: 'string'
       harness:
-        description: 'Harness to be used during "run" operation ("inference-perf", "guidellm", "inferencemax", "vllm-benchmark", "eval-containers", "nop")'
+        description: 'Comma-separated list of harnesses to be used during "run" operation ("inference-perf", "guidellm", "inferencemax", "vllm-benchmark", "eval-containers", "nop"). Paired positionally with "workload"; each pair is run against the same standup.'
         required: false
         default: 'inference-perf'
         type: 'string'
       workload:
-        description: 'Workload profile to be used during "run" operation (check list under "workload/profiles")'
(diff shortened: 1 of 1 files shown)


maugustosilva:
I wonder if the step `Upload results to GCS` also requires changes. We need to ensure the layout of what is pushed to the GCS `bucket` does not change.

Vezio:
@maugustosilva Good catch. The upload step now loops over every run workspace directory (multi-pair runs create one per run invocation) instead of capturing a single name. So now each directory is uploaded with the same `gcloud storage cp -r` call as before, so the `<date>/<run-dir>/results/...` bucket layout is unchanged

maugustosilva (approved):

Issue #1744, opened 2026-08-07
Inference-perf BR v0.2 conversion omits input_token_rate


danehans:
### Component

Analysis / Benchmark Report v0.2 conversion

### Description

The inference-perf converter in `native_to_br0_2.py` writes aggregate `output_token_rate`, `total_token_rate`, and `request_rate`, but does not write `input_token_rate`.

The Benchmark Report v0.2 schema supports all three token-rate fields. Consumers such as Prism therefore cannot display input token throughput from the generated report, even though inference-perf `summary_lifecycle_metrics.json` contains enough information.

### Current behavior

`results.request_performance.aggregate.throughput.input_token_rate` is absent from inference-perf BR v0.2 output.

### Expected behavior

The converter should populate `input_token_rate`. It can use a native input-token throughput value when available, or derive it as `total_token_rate - output_token_rate`.

The converter should preserve `None` when the required source values are unavailable and include regression coverage for the mapping.

### Reproduction

1. Run an inference-perf benchmark with analysis enabled.
2. Compare `summary_lifecycle_metrics.json` with the generated `benchmark_report_v0.2,_*.yaml`.
3. Observe that output and total token rates are present in the report, while input token rate is missing.

abhay1999:
saw this — looks like PR llm-d/llm-d-benchmark#1713 already fixed it, merged same day this was filed. `input_token_rate` is in `native_to_br0_2.py` on main now, reading from `successes.throughput.input_tokens_per_sec` same as described here. checked the v0.2.1 converter too, it just delegates to this one so it picks up the fix automatically.

can this be closed?

Bslabe123:
Confirmed, #1713 landed the mapping the same day this was filed and `input_token_rate` is populated on main, so the bug as described is fixed. Filed #1891 as the follow-up: it switches the inference-perf converter over to the BR v0.2.1 partials inference-perf now emits natively (kubernetes-sigs/inference-perf#461), which removes the local aggregate mapping that #1713's fix lives in. The upstream adapter emits `input_token_rate` directly, so the field is covered on that path too. Closing in favor of #1891.

PR #1911, opened 2026-09-12
feat: prefer inference-perf's native BR0.2 partial over hand-derived aggregate

### LINKED ISSUE #1891: Switch the inference-perf BR converter to inference-perf's native BR v0.2.1 partials
### Component

Analysis/Plotting

inference-perf now emits BR v0.2.1 partial reports natively (kubernetes-sigs/inference-perf#461, merged Aug 26): one file per stage, `inference-perf.partial.stage_<N>.yaml`, carrying `version`, `run` (uid, eid, time) and `results.request_performance.aggregate`, computed by the harness from the same request lifecycle metrics our converter reads.

`import_inference_perf` in `benchmark-report/llmd_benchmark_report/native_to_br0_2.py` re-derives that same block from `stage_*_lifecycle_metrics.json`: roughly 690 lines of latency, throughput and service mapping (native_to_br0_2.py:1946-2631) that now has an upstream equivalent. Two implementations of one mapping drift, and every new inference-perf metric costs a second patch here. #1744 is an instance: `input_token_rate` was missing on our side (fixed by #1713) while the upstream adapter has always emitted it (`inference_perf/reportgen/br/v0_2/adapter.py:137`).

### Proposed solution

Have the inference-perf path consume the partial when one sits next to the native results file:

1. `llmdbenchmark/analysis/scripts/inference-perf-analyze_results.sh` globs `stage_*.json`; teach the converter to look for the sibling `inference-perf.partial.stage_<N>.yaml`.
2. When it exists, take `results.request_performance.aggregate` and `run.uid` / `run.eid` / `run.time` from the partial and deep-merge onto the envelope we already compose (`_populate_benchmark_report_from_envars`, `_populate_run`, `_populate_load`, `_populate_stack`, and the `scenario.load.standardized` mapping). Partials are emitted with `None` fields stripped precisely so they merge this way.
3. Delete the now-unreachable aggregate construction, including the `input_token_rate` mapping #1713 added, keeping the native derivation only as the fallback for older harness images.
4. Keep fixture-based regression coverage, asserting the partial-sourced and natively derived reports agree field for field on the same run.

### What stays on the native path

- `results.session_performance` (`import_inference_perf_session`): not in the partial yet, tracked upstream as kubernetes-sigs/inference-perf#718.
- The v0.2.1 multimodal additions in `native_to_br0_2_1.py`: upstream kubernetes-sigs/inference-perf#719.
- Everything outside `results`: run identity, scenario and load, stack and components, and the observability block `embed_metrics.py` merges in after conversion. The partial deliberately carries no envelope.

So this replaces the aggregate half of `import_inference_perf`, not the whole importer.

### Sequencing

`build/Dockerfile` pins `INFERENCE_PERF_BRANCH`/`INFERENCE_PERF_COMMIT` to `v0.6.1`, which predates #461, so no partial exists in the harness image today. This lands after that pin moves to an inference-perf release containing the partials (v0.7.0, tracked in kubernetes-sigs/inference-perf#606). Detecting the partial rather than requiring it means the converter degrades to today's behavior until then.

### Alternatives

Keep patching the local mapping field by field, as in #1744 and #1713. Cheap per change, but it keeps two copies of one computation and consumers see different field sets depending on which producer they read.

### Additional context

Relates to #1744; once this lands, the mapping added by #1713 is dead code on the partial path.

elinacse:
Fixes #1891

Co-Authored-By: Claude Sonnet

inference-perf >= v0.7.0 (kubernetes-sigs/inference-perf#461) writes
inference-perf.partial.stage_<N>.yaml next to each stage_<N>_lifecycle_
metrics.json, carrying results.request_performance.aggregate and
run.{uid,eid,time} computed from the same request lifecycle metrics
import_inference_perf otherwise re-derives by hand in ~700 lines
(native_to_br0_2.py, previously inline in the function). Two
implementations of one mapping drift -- #1744 (missing input_token_rate)
was an instance of exactly that.

- Add _find_inference_perf_partial: looks for the sibling partial next to
  results_file and returns its parsed contents, or None (missing or
  unparseable) so the native path is unaffected.
- When a partial is found, take only run.uid/eid/time (present fields win
  over the envelope's placeholders -- run.uid is explicitly "Initial UID,
  may be updated"; run.time here is the stage's own wall-clock window,
  more precise than the overall-harness window every stage previously
  shared) and results.request_performance.aggregate verbatim. Nothing
  else in the partial (version, scenario, stack) touches the envelope
  import_inference_perf already composes.
- Extract the existing native aggregate construction, unchanged, into
  _build_inference_perf_aggregate_native as the fallback for harness
  images built from an inference-perf release that predates the partial.

What still runs the native derivation, unchanged: results.session_performance
(import_inference_perf_session -- no partial yet, kubernetes-sigs/
inference-perf#718) and the v0.2.1 multimodal additions in
native_to_br0_2_1.py (request_size, multimodal, media rates --
kubernetes-sigs/inference-perf#719).

Added tests/test_inference_perf_br_partial.py: the partial-sourced and
natively-derived reports agree field-for-field on the same run, a
distinct partial aggregate is actually used (not silently ignored),
partial lookup is scoped to its own stage, and both the no-partial and
malformed-partial cases fall back to the native derivation.

### DIFF
--- a/benchmark-report/llmd_benchmark_report/native_to_br0_2.py
+++ b/benchmark-report/llmd_benchmark_report/native_to_br0_2.py
@@ -1803,6 +1803,30 @@ def import_eval_containers(results_file: str) -> BenchmarkReportV02:
     return load_benchmark_report(br_dict)
 
 
+def _find_inference_perf_partial(results_file: str, stage: int) -> dict | None:
+    """Return the sibling BR0.2 partial for this stage, if inference-perf wrote one.
+
+    inference-perf >= v0.7.0 (kubernetes-sigs/inference-perf#461) drops
+    ``inference-perf.partial.stage_<N>.yaml`` next to each
+    ``stage_<N>_lifecycle_metrics.json``, carrying ``results.request_performance
+    .aggregate`` and ``run.{uid,eid,time}`` computed from the same request
+    lifecycle metrics this module otherwise re-derives by hand. Preferring the
+    partial when present retires that duplicate ~700-line mapping (llm-d/
+    llm-d-benchmark#1891); older harness images with no partial fall back to
+    the native derivation below unchanged.
+    """
+    partial_path = os.path.join(
+        os.path.dirname(results_file), f"inference-perf.partial.stage_{stage}.yaml"
+    )
+    if not os.path.isfile(partial_path):
+        return None
+    try:
+        return import_yaml(partial_path)
+    except (OSError, yaml.YAMLError) as e:
+        sys.stderr.write(f"Failed to read inference-perf partial {partial_path}: {e}\n")
+        return None
+
+
 def import_inference_perf(results_file: str) -> BenchmarkReportV02:
     """Import data from a Inference Perf run as a BenchmarkReportV02.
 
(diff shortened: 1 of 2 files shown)


elinacse:
@Vezio Could you please review this PR ? Thanks !

benchmark-report/llmd_benchmark_report/native_to_br0_2.py
```diff
@@ -1943,6 +1967,48 @@ def import_inference_perf(results_file: str) -> BenchmarkReportV02:
...
+        # overall-harness window every stage would otherwise share).
+        run_from_partial = {
+            k: v
+            for k, v in (get_nested(partial, ["run"], {}) or {}).items()
+            if k in ("uid", "eid", "time")
+        }
+        if run_from_partial:
+            update_dict(br_dict, {"run": run_from_partial})
+        aggregate = get_nested(partial, ["results", "request_performance", "aggregate"])
+    else:
```
ruocco, line 1988: Can you change the `else` with `if aggregate is None` ? This way if get_nested returns None it still falls back to `_build_inference_perf_aggregate_native`

Vezio (approved):

Vezio:
Once CI Passed we can merge this

ruocco (changes requested):
@elinacse thank you for this, I think this should fix a latent bug I was observing when having prism read trace_replay "session" reports from the PVC.

benchmark-report/llmd_benchmark_report/native_to_br0_2.py
Vezio: @elinacse once you make ^ we can merge

Bslabe123:
Can we hold this until the inference-perf pin moves to v0.7.0? `build/Dockerfile` still pins v0.6.1, which doesn't write partials, so the new branch can't run yet. Once the pin moves, this should also delete `_build_inference_perf_aggregate_native` rather than keep it as a fallback, since removing that duplicate mapping is the point of #1891.

A heads up too, inference-perf v0.7.0 is due to be cut today.

Bslabe123:
^update, v0.7.0 cut

PR #1912, opened 2026-09-12
fix: skip the Kubernetes template set on nok8s stacks

### LINKED ISSUE #1704: nok8s: the no-Kubernetes path renders the full Kubernetes manifest set and warns three times that helm/skopeo are missing

elinacse:
_load_templates() rendered every non-partial template unconditionally, so a nok8s stack still got the full Kubernetes manifest set written to disk -- PVCs, RBAC, the harness pod, helmfiles, HTTPRoute, PodMonitor -- 36 files per stack, none of which nok8s ever applies (only step_05_nok8s_deploy.py/step_06_nok8s_teardown.py read back the four 31-34_nok8s-* templates).

Skip any non-nok8s-named template when is_nok8s is set, at the per-stack render loop rather than the (globally cached) template loader, so a mixed k8s/nok8s scenario renders correctly per stack. k8s stacks are unaffected: is_nok8s is false, so every template still renders exactly as before.

Fixes #1704 (the remaining half after #1758 fixed the false helm/skopeo warnings)

### DIFF
--- a/llmdbenchmark/parser/render_plans.py
+++ b/llmdbenchmark/parser/render_plans.py
@@ -47,6 +47,15 @@ class RenderPlans:
     # Prefix for partial/macro files (not rendered directly)
     PARTIAL_PREFIX = "_"
 
+    # Substring marking a template as nok8s-specific (e.g. 34_nok8s-containers.yaml.j2).
+    # On a nok8s stack, only these are ever read back (see step_05_nok8s_deploy.py
+    # and step_06_nok8s_teardown.py): everything else is a Kubernetes manifest --
+    # PVCs, RBAC, the harness pod, helmfiles, HTTPRoute, PodMonitor -- that nok8s
+    # never applies. Rendering them anyway produced 30+ dead files per stack and
+    # version-resolver warnings for tools (helm, skopeo) the nok8s path itself
+    # tells users are unnecessary (docs/nok8s.md) (#1704).
+    NOK8S_TEMPLATE_INFIX = "nok8s"
+
     # Default namespace when "auto" is specified (matches original bash: llmdbench)
     DEFAULT_NAMESPACE = "llmdbench"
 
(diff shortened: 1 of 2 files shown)


elinacse:
Hi @Vezio Can you please review this PR ? Thanks !

mengmeiye (approved):

PR #1918, opened 2026-09-14
Fix/1100 reset caches

### LINKED ISSUE #1100: Evict the cache between repeated benchmark invocations
### Component

Run

The current implementation of the repeat flag for `run_only` doesn't clear the cache between benchmark iterations. As a result, previous runs affect new runs, and the experiments become dependent on each other.

### Proposed solution

Evict the kv cache (all tiers) between benchmark iterations. I'm not sure if this is already supported by vllm/llm-d and just needs to be added to the repeat feature implementation, or requires additional lower levels support.

### Alternatives

_No response_

Related PR: https://github.com/llm-d/llm-d-benchmark/pull/859

abhisekupadhyaya:
Closes #1100

`reset_caches: true` (from #1616) resets the vLLM caches before each treatment group, but two things were missing:

1. Offload tiers were not reset. The reset URL had no query string ([endpoint.py#L951-L952](https://github.com/llm-d/llm-d-benchmark/blob/2abdb608c4e64fccbabfe171298a88471e2cadc4/llmdbenchmark/utilities/endpoint.py#L951-L952)), and vLLM only resets connector-managed caches when `reset_external=true` is passed. With CPU offloading the next treatment still started warm.

2. A refused reset looked like a success. Since v0.26.0 vLLM answers `/reset_prefix_cache` with HTTP 200 and `{"success": false}` while blocks are still held, e.g. by in-flight offload transfers (vllm-project/vllm#46893). We only checked the status code ([endpoint.py#L1001-L1004](https://github.com/llm-d/llm-d-benchmark/blob/2abdb608c4e64fccbabfe171298a48471e2cadc4/llmdbenchmark/utilities/endpoint.py#L1001-L1004)). This is not a bug in #1616: the success body did not exist yet when it merged, and the pin moved to v0.26.0 in #1747.

What this PR does:

- Always passes `reset_external=true` on the prefix-cache reset. Older vLLM ignores the parameter, and it is a no-op without a KV connector. The endpoint and URL are quoted in the curl loop because of the `?`.
- Parses each reply body. A 200 with `{"success": false}` marks that pod as refused; an empty body (vLLM < 0.26) is treated as success, same as before.
- Retries only the refused pods, only on the prefix-cache endpoint, up to `max_retries=3` attempts in total with `retry_interval=5` seconds between them. A reset fires between treatments, so the budget only needs to outlast in-flight offload transfers. Names follow `test_model_serving` in the same module.
- Logs each retry, so a pause in the run log is explained.
- If a pod is still unconfirmed after that, logs one warning naming the pod and the attempt count, as agreed on the issue. Resets stay non-fatal by default.
- Adds a top-level `reset_caches_required` key (default false). When set, a reset that cannot be confirmed aborts the run before that group starts, since nothing is running yet. Any reset warning counts, a 404 from dev mode being off included. YAML only, like the other run controls. As @ruocco suggested for experiments that are meaningless without a cold cache.

Verification:

- 17 unit tests: 12 in `tests/test_endpoint_reset.py` (7 existing, 5 new, including two pods where only one refuses) and 5 for the flag across `test_deploy_harness_status.py`, `test_treatment_groups.py` and `test_run_controls.py`.
- An end-to-end run of the generated shell script with real `sh` and `curl` against a small local HTTP server that mimics vLLM 0.26: no refusal, one refusal then success, and permanent refusal all behaved as expected.
- Not run on a GPU cluster. I do not have one; if someone can run one of the `reset_caches: true` experiments on a real stack, `vllm:external_prefix_cache_hits_total` on the first requests of a treatment would show whether the offload tier is now cold.

Not covered here:

- LMCache is still not cleared. vLLM's LMCache connector does not implement `reset_cache()`, so vLLM reports success without clearing it, and `reset_caches_required` cannot see through it either. That needs a fix on the vLLM side.
- A treatment retry does not re-fire the reset. The comment at step_07's retry loop says it does, but the reset only runs once per group, so a retried attempt starts warm and the new flag cannot catch it. Left as is here since changing it affects every run; happy to follow up.

Related: llm-d-incubation/llm-d-skills#43 adds the same `success: false` handling to the clear-cache skill script.


Vezio:
Please sign your commits.

abhisekupadhyaya:
Signed now, thanks, and the signing key is registered. Also took the Copilot note: reset_caches_required without reset_caches now fails the run instead of only warning. The signed-commits check needs a re-run, and the other workflows are waiting for approval since this is my first PR here.

abhisekupadhyaya:
@Vezio It is signed now

abhisekupadhyaya:
@Vezio @achandrasekar @kalantar  @maugustosilva @mengmeiye @namasl
Does the PR look good?

maugustosilva (approved):

PR #1935, opened 2026-09-17
Fix nested guide path resolution for multimodal-serving scenarios

rlakhtakia:
Fix for Issue 3 in https://github.com/llm-d/llm-d/issues/2511. 

This PR fixes the CI failures caused by the new nested multimodal nightly workflows.

Changes:

1. Handle nested paths: Updated resolve_specification_file to translate slashes into hyphens (e.g., guides/multimodal-serving/aggregation → guides/multimodal-serving-aggregation) so the CLI can find the configuration files.

2. Fix Helm release names: Updated GuideVariableResolver to strip slashes from GUIDE_NAME (Helm doesn't allow slashes in release names).

3. Fix Kustomize paths: Updated GuideVariableResolver to split compound strings like e-pd/gke into TOPOLOGY=e-pd and INFRA_PROVIDER=gke. This fixes the evalsymlink failure caused by duplicated paths.

4. Add missing files: Added the missing .yaml.j2 templates and the e-disaggregation scenario file needed by the CI.


llmdbenchmark/kustomize/variable_resolver.py
```diff
@@ -68,6 +68,18 @@ def __init__(
...
 
+        # Strip path prefixes from GUIDE_NAME
+        if "/" in guide_name:
+            guide_name = guide_name.split("/")[-1]
+
+        # Split INFRA_PROVIDER into TOPOLOGY and INFRA_PROVIDER
+        if variable_overrides and "INFRA_PROVIDER" in variable_overrides:
+            infra = variable_overrides["INFRA_PROVIDER"]
+            if "/" in infra:
+                topology, actual_provider = infra.split("/", 1)
```
achandrasekar, line 79: Is this backward compatible with the other guides as well?

llmdbenchmark/kustomize/variable_resolver.py
```diff
@@ -68,6 +68,18 @@ def __init__(
         self._repo_path = Path(repo_path).resolve()
         self._accelerator_backend = accelerator_backend
 
+        # Strip path prefixes from GUIDE_NAME
+        if "/" in guide_name:
+            guide_name = guide_name.split("/")[-1]
```
achandrasekar, line 73: Can it lead to collision between two different guides like `multi-modal-serving/pd` and `tiered-prefix-cache/pd` because we are stripping the prefix? Should we use the hyphenation approach below for this too to flatten the name instead?

config/scenarios/guides/multimodal-serving-e-disaggregation.yaml
```diff
@@ -0,0 +1,392 @@
+# MULTIMODAL OPTIMIZED BASELINE (AGGREGATION) WELL LIT PATH for LLM-D Deployment
+# Migrated from bash environment variables to YAML format
```
achandrasekar, line 2: For my understanding, where were these env variables specified before? In this repo or directly in the Github action? Also, how can we prevent future breakages like this - should we enforce something on the guide owners?

achandrasekar:
Looks like lint and format check seems to be failing as well. Not sure if it is related to your changes.

llmdbenchmark/kustomize/variable_resolver.py
rlakhtakia: They are deployed in different namespaces, which would provide strict isolation for services and labels (and CI dynamically generates a unique namespace for every run). If we used the hyphenated approach, we'd have to do a massive refactor of all the hardcoded matchLabels across the llm-d manifests to match.

llmdbenchmark/kustomize/variable_resolver.py
rlakhtakia: yes!! ran optimized baseline against these changes and verified it worked.

config/scenarios/guides/multimodal-serving-e-disaggregation.yaml
rlakhtakia: I believe they are in the README files. This is just the standard across the guides.

llmdbenchmark/kustomize/variable_resolver.py
achandrasekar: Makes sense.

achandrasekar (dismissed):

achandrasekar:
Looks like one of the commits is missing the signature.

achandrasekar (approved):

